PR #2, opened 2019-08-16
Startin UI with Landing and Project Index

wgranger:
This PR begins the initial layout of the app. The landing page should be visible when loading the app, and the `/projects` route should also be set up. This is very basic, and I chose this as a stopping point before the code got too bloated so I could get feedback on how things are set up in Grommet before proceeding further. I use mock data here.


src/components/IndexTable/IndexTable.js
```diff
@@ -0,0 +1,72 @@
...
+                />
+              ),
+              header: (
+                <CheckBox
+                  checked={checked.length === data.length}
+                  indeterminate={
+                    checked.length > 0 && checked.length < data.length
+                  }
+                  onChange={this.onCheckAll}
+                />
```
srallen: These checkboxes need labels for accessibility. There's a prop for `CheckBox` that can be used or you can use `htmlFor` and `id` to pair up a label. Maybe the workflow name is the label? It depends on what these checkboxes are supposed to do.

src/components/IndexTable/IndexTable.js
```diff
@@ -0,0 +1,72 @@
+import React, { Component } from 'react'
+import { Box, CheckBox, DataTable } from 'grommet'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+const RoundedBox = styled(Box)`
+  border-radius: 0.5em;
+`
+
+class IndexTable extends Component {
```
srallen: Is this going to be shared between the projects list, workflows list, etc? Perhaps name this `ResourcesTable` or something a bit more descriptive? If it's not shared, then specifically what it is `ProjectsTable`?

srallen:
@wgranger the Zooniverse Grommet v2 theme is available on npm at version 2.1.0. `yarn add @zooniverse/grommet-theme` and then there's an example on use in its readme on the repo.

src/components/Badge/Badge.js
```diff
@@ -0,0 +1,32 @@
...
+
+const CapitalText = styled(Text)`
+  line-height: 0.5em;
+  text-transform: uppercase;
+`
+
+function Badge (props) {
+  return (
+    <Box direction='row' width='14em'>
+      <StyledAvatar fit='contain' src={DefaultAvatar}/>
```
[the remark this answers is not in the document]
wgranger: I'm unable to get Grommet's `Image` component to accept rounded borders, which is why I've resorted to using an image tag with styled components. The monorepo seems to use a [similar solution ](https://github.com/zooniverse/front-end-monorepo/blob/0646cd983059a46e5e11eaf1cfb2359b15f01b4a/packages/app-project/src/components/ProjectHeader/components/Avatar/Avatar.js#L10).

wgranger:
I believe I'd addressed all comments from the initial review. This should be good for another look.

PR #8, opened 2019-08-29
Introduce HeaderButton Component and Containers

wgranger:
This PR introduces a `HeaderButton` functional component to render all buttons to be used in the header. I am using Font Awesome again because grommet-icons is lacking and I the package allows me to rotate icons with a prop (needed for the layout button). 

All buttons are visible in the storybook `yarn storybook`

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,7 +1,7 @@
 import { configure } from '@storybook/react';
 
  // automatically import all files ending in *.stories.js
-const req = require.context('../stories', true, /.stories.js$/);
+const req = require.context('../src', true, /.stories.js$/);
 function loadStories() {
   req.keys().forEach(filename => req(filename));
 }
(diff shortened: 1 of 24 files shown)


srallen:
Is the design actually using font awesome icons or can we grab the svg path and make a custom one? Grommet icons support customization and I can link to an example of us doing it in the monorepo.

wgranger:
The design uses font awesome for items like the save button and icons in the layout button

wgranger:
No, this is supposed to be based on the inVision design, which is a bit different from the basic Grommet checkbox. https://projects.invisionapp.com/d/?origin=v7?origin=v7#/console/17925240/378388615/preview

shaunanoordin:
I don't have access to that inVision file unfortunately, but if the design is as intended, then cool! 👍

wgranger:
The oddly-sized toggle box should now be taken care of.

PR #9, opened 2019-09-05
Add Subject Viewer Components

wgranger:
I created a good bit of files here to introduce the foundation of the subject viewer, which is available at the subjects/editor route as well as the storybook.


wgranger:
Good catch. The most recent commit should take care of this.

shaunanoordin:
👍 Noice! Those buttons are looking good.

PR #10, opened 2019-09-09
Add Aggregated Transcriptions with Mock Data

wgranger:
This PR adds layout for the aggregated transcriptions data table. You can view these changes through a storybook. 

I'll need to rebase this and move some files around after a couple PRs merge to follow how to code is being organized.


wgranger:
@beckyrother Do you have a preference on the comment above? Cutting off text with ellipses vs. wrapping onto the next time.

beckyrother:
This is a really good question @shaunanoordin. Initially it had been intended to just truncate and you'd have to click into the details to read the full line, but it absolutely makes sense that someone would want to read the document at a glance. 
Let's wrap the text onto the next line, and let the dots/data on the right side align with the last line of text (leaving a bit of white space above those data).
Thanks!

wgranger:
Alright. Think I've taken care of that. Styling grommet cells in a data table is tricky though.

PR #12, opened 2019-09-20
Route to Classifier

wgranger:
Do not be intimidated by the number of lines added! They are mainly boilerplate tests.

[Link to InVision prototypes](https://projects.invisionapp.com/d/?origin=v7?origin=v7#/projects/prototypes/17925240)

This PR does several things:

- Routes from the landing page to the classifier. Click "log in" to proceed to the projects page. Login isn't hooked up yet, so no need to complete the form. I'd like some feedback on the routing. I was thinking of doing the projects/:id/workflows/:id/subject-sets/:id, etc. type routing, but I'm not sure if this is necessary.

- Restyle some items after viewing components on their respective page.

- Add render tests to many components. I did this as one of the final commits to bring the code coverage up, but it mainly consists of checks that the rendered component doesn't crash.


src/screens/ProjectsIndex/mockProjects.js
```diff
@@ -0,0 +1,59 @@
+const personalProjects = [
+    {
+      src: '',
+      title: 'Anti-Slavey Manuscripts'
```
shaunanoordin: _Anti-Slavey?_ 🤔

wgranger:
Cool! I fixed the Slavey typo and linked to "edit" instead of "editor," although those paths may change in the future. No luck on the _major_ one. I'm just all out of options on how else to spell it.

I'll keep the note on folder organization in mind going forward. I suppose using sub-directories is only necessary when many extra files or components are present.

PR #13, opened 2019-09-27
Allow Login

wgranger:
Login is working on this branch with tests with one exception. I'm able to login and logout seemingly without problems. However, when you reload the page again after logging out, you'll see there is still a user logged in. Thoughts? Aside from that, this PR is ready for review and hooks up the first components with mobx and hooks.


camallen:
> However, when you reload the page again after logging out, you'll see there is still a user logged in. Thoughts? 

I cannot reproduce this behaviour, logging in gets me a session & cookie on `panoptes-staging.zooniverse.org` and then token is returned via this session cookie.

I then log out and reload the page and receive a 401 on the `/oauth/token` req and am logged out both locally and on the panoptes-staging.zooniverse.org API.

I've tested locally on this branch:
```
# node v8.9.3
git checkout -b allow-login origin/allow-login
npm install
npm start
```

wgranger:
@camallen Thanks for also taking a look at this. I cleared my cookies and everything appears to be working now. It might've been an issue with having multiple Panoptes cookies set.

@shaunanoordin This should be ready for a full review since rebasing this with master. You shouldn't have any issues with login/logout as mentioned in the opening comment above.

wgranger:
I hadn't considered adding the LoginForm and Badge into storybook yet. There's no real reason why they aren't, but perhaps I can add this next time I change the components so they can be viewed in isolation.

PR #15, opened 2019-10-01
Add SearchModal

wgranger:
This PR adds the SearchModal components while also allowing the form to function using Formik and Mobx


wgranger:
As always, thanks for the thorough review. I'll address these ❓questions before merge so I don't have to come back to this in the future. I'll try to get this taken care of today.

wgranger:
Form validation should now be working properly, but please let me know if there is anything left outstanding.

beckyrother:
Completely agree, and I like the error message but I'd modify it to "Select an ID type" to be a bit more specific and help people find what they're supposed to select

wgranger:
👍 Just updated validations based on our discussion.

shaunanoordin:
New validation updates work great. 👍 LGTM!

PR #16, opened 2019-10-03
Add Image Tool Functionality

wgranger:
This pr adds functionality for all image tools you can view by hovering over a subject. You can test these in storybook. I also added a refresh button and tests.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,0 +1,10 @@
+const path = require('path')
+
+module.exports = (baseConfig, env, config) => {
+  config.resolve.modules = [
+    ...(config.resolve.modules || []),
+    path.resolve(__dirname, "../src")
+  ];
+
+  return config;
+}
--- a/src/components/SubjectViewer/SubjectViewer.js
+++ b/src/components/SubjectViewer/SubjectViewer.js
@@ -13,17 +13,19 @@ const G = styled.g`
 
 function SubjectViewer ({ ref, rotation, scale, translateX, translateY, url }) {
   const transform = `scale(${scale}) translate(${translateX}, ${translateY}) rotate(${rotation})`
+  const inputEl = React.useRef(null);
+  const boundingBox = (inputEl && inputEl.current && inputEl.current.getBoundingClientRect());
 
   return (
-    <SVG ref={ref} viewBox="0 0 100 100">
+    <SVG ref={inputEl}>
       <G transform={transform}>
         <image
           height='100%'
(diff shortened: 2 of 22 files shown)


shaunanoordin:
I'm not sure if this branch is functioning, as any attempt to run `yarn storybook` results in an error.

- After deleting `node_modules` and re-running `yarn install`...
- ...running `yarn storybook` results in errors like" `ERROR in ./src/components/SubjectViewer/SubjectViewerContainer.js
Module not found: Error: Can't resolve 'store' in '/Users/shaun/projects/text-editor/src/components/SubjectViewer'` in both the bash console and in `http://localhost:6006/`
- Note: I can still view the normal web app, e.g. `yarn start` -> `http://localhost:3000/projects/workflows/subject-sets/subjects/edit`, just fine.

wgranger:
Ahh right. Storybook is crashing on the absolute imports I have for the store. Let me see if I can find a workaround. I failed to notice this crashed storybook when completing this branch.

eatyourgreens:
If you’re using webpack aliases, you’ll have to pass them to the storybook. https://github.com/zooniverse/front-end-monorepo/pull/1165

shaunanoordin:
I'm not familiar enough with Storybook to suggest a specific solution. However, if it's based on `webpack`, the latest versions allow for `alias` paths using [`modules.exports.resolve.alias = {...}`](https://webpack.js.org/configuration/resolve/)

So you can do something like...

```
// in webpack...
module.exports = {
  resolve: {
    alias: {
      'MyStoreDirectory': path.resolve(__dirname, 'src/stores/'),
     }
  }
}

// in your component...
import store from `MyStoreDirectory`
```

shaunanoordin:
What Jim said! 😄☝️

eatyourgreens:
If you’re testing with mocha, you might also need the mocha config changes from that PR.

wgranger:
Thanks for the comments! I had to use a slightly different approach since I'm limited to webpack configuration with create-react-app. I put the following into `webpack.config.js` in the `.storybook` folder.

```
const path = require('path')

module.exports = (baseConfig, env, config) => {
  config.resolve.modules = [
    ...(config.resolve.modules || []),
    path.resolve(__dirname, "../src")
  ];

  return config;
}
```

wgranger:
Also, several of these tests will fail until I can get these changes into the panoptes-client. You'll have to monkey patch for now.

https://github.com/zooniverse/panoptes-javascript-client/pull/119

PR #17, opened 2019-10-03
Add Header Tools by Path

wgranger:
This PR changes which header tools are shown depending on the path given. I've also added tests for the components I touched. I've help off adding functionality here using a mobx store until #16 is merged with the `ImageStore`. I'll add button functions in a separate PR.

[Edit page layout](https://projects.invisionapp.com/d/?origin=v7?origin=v7#/console/17925240/371813754/preview)
[Subjects Page layout](https://projects.invisionapp.com/d/?origin=v7?origin=v7#/console/17925240/375544496/preview)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,6 +12,13 @@ import WorkflowsIndex from './screens/WorkflowsIndex'
 import Editor from './screens/Editor'
 import Header from './screens/Header'
 import { mergedTheme } from './theme'
+import {
+  PROJECTS_PATH,
+  WORKFLOWS_PATH,
+  SUBJECT_SETS_PATH,
+  SUBJECTS_PATH,
+  EDIT_PATH
+} from 'paths'
 
 function App() {
   const store = React.useContext(AppContext)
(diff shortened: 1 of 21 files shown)


wgranger:
Alright, good points above. I rigged the app to accept routes with parameters (eg. `/projects/123/workflows`) and found out `react-router-dom` has a `matchPath` helper for determining if a path matches a route, so no need to do any regex or worry about capitalization woes. I think this should be a suitable solution, and tests are still passing.  

Also, I added `wrap` onto the header tools box, so that should take care of the messy header, but let me know if there are still any outstanding issues.

shaunanoordin:
👍 Noice! The path-matching changes look good in the code; I'll test this when I'm back from today's London expedition.

(But if you can't wait to merge, then I think the latest changes LGEnoughTM. 👌)

shaunanoordin:
## PR Review (Update)

Righto, I've checked the updates 👌 

- The new full paths look and function well 👍 - e.g. `http://localhost:3000/projects/123/workflows/123/subject-sets/123/subjects/123/edit` now replaces `http://localhost:3000/projects/workflows/subject-sets/subjects/edit` and the code does path matching quite well.
- The header tools still seem to not wrap properly on small viewports, though.

I'm still seeing this...
Screen Shot 2019-10-08 at 15 58 51

...when I'm expecting...
Screen Shot 2019-10-08 at 15 59 07

That second screenshot was achieved by adding a 'wrap' here...

```
export default function EditorHeader ({ buttons }) {
  return (
    <Box as='header' direction='row' pad={{ bottom: 'small' }} border='bottom'>
      <Box direction='row' justify='between' border='right' fill='horizontal'>
        <Box align='center'>
          <StyledHeader level='2'>Text Transcription Project</StyledHeader>
        </Box>
        <Box align='center' direction='row' gap='small' pad={{ horizontal: 'medium' }} wrap>  // <-- HERE
          {buttons.map((HeaderButton, i) => <HeaderButton key={`HEADER_BUTTON_${i}`} />)}
        </Box>
      </Box>
      <Badge />
    </Box>
  )
}
```

However, since you mentioned _"I added wrap onto the header tools box",_ I'm not sure if you perhaps added it already, but haven't pushed the commit?

wgranger:
Oops. Yes, that's exactly what happened. I just forgot to push that commit. My change was the one as you described above. Sorry for the confusion.

PR #20, opened 2019-10-10
Add Classifier Layout Controls

### LINKED ISSUE #34: Allow Classifier to Resize Viewers
The classifier has two viewers (subject and transcription) that should be able to resize on the click and drag of a central ellipsis icon. Resizing should occur in both horizontal and vertical layout modes.

wgranger:
This PR adds two controls to the classifier. 

- Allow users to switch between landscape modes using the header button
- Allow users to change widths and heights of the subject and transcription viewers using the ellipsis drag handler.

[Editor page design](https://projects.invisionapp.com/d/?origin=v7?origin=v7#/console/17925240/371813754/preview)

Closes #34

### DIFF
--- a/src/components/AggregatedTranscriptions/AggregatedTranscriptions.js
+++ b/src/components/AggregatedTranscriptions/AggregatedTranscriptions.js
@@ -26,7 +26,7 @@ const OverflowBox = styled(Box)`
 
 function AggregatedTranscriptions (props) {
   return (
-    <Box background='white' height='large' round='xsmall' width='large'>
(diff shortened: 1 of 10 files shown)


wgranger:
Nice! Good suggestions. I've renamed all "classifier" references to "editor." I suppose my brain naturally defined that area as a classifier when that is not at all what is happening.

I also added in the `e.preventDefault()` but I didn't find adding `stopPropagation `, `return false`, or any of the other function calls was necessary, unless there's something happening under the hood I'm unaware of.

PR #35, opened 2019-10-11
Setup docker

### LINKED ISSUE #14: add readme explaining how to get started with this app
Flesh out the readme to explain how to run tests and smoke test this application from scratch. 

ideally it'll use docker to build the app / deps and then docker-compose to run the app locally with all it's dependencies built it.

wgranger:
This PR sets up docker on the repo to quickly get the app up and running. With this PR, `docker-compose up` will run a dev version of the app and storybook together. I've also updated the README to reflect the new commands, which closes a recent issue.

Closes #14

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM node:10
+
+RUN mkdir -p /usr/src
+WORKDIR /usr/src/
+
+COPY ./ /usr/src
+RUN chown -R node:node .
+
+USER node
+
+RUN yarn install
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,3 +14,13 @@ services:
     volumes:
       - ${PWD}:/src
       - /src/node_modules
+  storybook:
+    image: zooniverse-text-editor:latest
+    entrypoint:
+      - "yarn"
+    command: ["storybook"]
+    ports:
+      - "6006:6006"
+    volumes:
+      - ${PWD}:/src
+      - /src/node_modules
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,16 @@
+version: '3.4'
+
+services:
+  dev:
+    image: zooniverse-text-editor:latest
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    entrypoint:
+      - "yarn"
+    command: ["start"]
+    ports:
+      - "3000:3000"
+    volumes:
+      - ${PWD}:/src
+      - /src/node_modules
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+**/build
+**/dist
+**/node_modules
+**/.DS_Store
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN mkdir -p /usr/src
+WORKDIR /usr/src/
+
+COPY ./ /usr/src
+RUN chown -R node:node .
+
+USER node
+
+RUN yarn install
```
eatyourgreens: You can make the Docker build much faster by installing your dependencies before copying the source code across. Something like
```docker
WORKDIR /usr/src

RUN chown -R node:node /usr/src
USER node

ADD ./package.json /usr/src
ADD ./yarn.lock /usr/src
RUN yarn install

COPY ./ /usr/src
```

That will speed up your build by [making use of Docker's cached layers](http://bitjudo.com/blog/2014/03/13/building-efficient-dockerfiles-node-dot-js/).

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+    build:
+      context: ./
+      dockerfile: Dockerfile
+    entrypoint:
+      - "yarn"
+    command: ["start"]
+    ports:
+      - "3000:3000"
+    volumes:
+      - ${PWD}:/src
```
eatyourgreens: This mounts the all of your local files to `/src` in the image, overwriting any files that were copied in during the build. The docker build also copied everything to `/usr/src` not `/src`.

I think what you want to do is mount your local `/src` directory to the image, which would be something like `- ${PWD}/src:/usr/src/src`.

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+      - /src/node_modules
```
eatyourgreens: You can get rid of this.

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+  storybook:
+    image: zooniverse-text-editor:latest
+    entrypoint:
+      - "yarn"
+    command: ["storybook"]
+    ports:
+      - "6006:6006"
+    volumes:
+      - ${PWD}:/src
+      - /src/node_modules
```
eatyourgreens: Same comments here about mounting volumes from your local filesystem to the container.

wgranger:
Thanks for the review! I believe I made the appropriate changes based on suggestions.

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+  storybook:
+    image: zooniverse-text-editor:latest
+    entrypoint:
+      - "yarn"
+    command: ["storybook"]
+    ports:
+      - "6006:6006"
+    volumes:
+      - ${PWD}/src:/usr/src/src
+      - /src/node_modules
```
eatyourgreens: You can get rid of `/src/node_modules` here too.
```suggestion
      - 
```

PR #37, opened 2019-10-16
Fetch and Display Sample Image

### LINKED ISSUE #24: Filmstrip Viewer Functionality
Flesh out filmstrip viewer implementation. This mainly deals with displaying all images of a multi-image subject However, there are a few lingering questions to address here:  

1. How exactly should we deal with the image rotation?
      - Can an image only be rotated by 90 degree increments?
      - Are we only concerned with a 90 degree rotation, left or right?
2. What should we do UI-wise if a subject has 12 pages?
3. Allow clicking on a page to change the image in the subject viewer

### LINKED ISSUE #27: Display Subject and Associated Data
Retrieve a subject from the API and display that subject in the designated place. The steps here will be:
- Choose a subject via the data table
- Display that subject on the subject viewer and filmstrip viewer
- Display subject metadata in the metadata modal

wgranger:
This PR handles several items on the editor page. Most importantly, it handles retrieving a subject image from Panoptes when selecting an item on the subjects data table and handles loading the editor page through a subject param. Note: the subject store currently loads a predetermined id of a known subject. 

Aside from this, this PR handles several other items.
- Display the subject on the editor
- Display the subject on the filmstrip viewer
- Allow switching between pages on the opened filmstrip viewer
- Allow switching between pages on the collapsed filmstrip viewer
- Display subject metadata via an "i" button press.

WIP as I clean up the code a bit.

Closes #24 Closes #27


wgranger:
Removed the WIP to this PR. I've added tests and fleshed out the components a bit. There are still tests missing for those components that use the `withRouter` wrapper and the new `SubjectStore`, but I intent to add those later, time permitting.

src/store/SubjectStore.js
```diff
@@ -0,0 +1,44 @@
...
+}).actions(self => ({
+  changeIndex: (index) => {
+    self.index = index
+  },
+
+  fetchSubject: flow (function * fetchSubject (id) {
+    self.asyncState = ASYNC_STATES.FETCHING
+    try {
+      const response = yield subjects.get({ id: TEMPORARY_SUBJECT_ID })
+      if (response.body.subjects[0]) {
```
shaunanoordin: I assume at this stage of development, we're not worried about creating a Subject queue yet. This is fine 👌

wgranger: No. I'm not sure a subject queue would be necessary since an editor always has to select a specific subject to edit from an index.

src/screens/SubjectsIndex/SubjectsPageContainer.js
```diff
@@ -1,14 +1,24 @@
...
-  render() {
-    return (
-      <Box margin='medium' fill='vertical'>
-        <ResourcesTable columns={mockColumns} data={mockData} />
-      </Box>
-    )
+function SubjectsPageContainer (props) {
+  const store = React.useContext(AppContext)
+  const onSelection = (id) => {
+    store.subject.fetchSubject(id, true)
```
shaunanoordin: `fetchSubject()` is called in two areas:
1. on the Subject Viewer/Editor page (showing a single Subject, i.e. http://localhost:3000/projects/123/workflows/123/subject-sets/123/subjects/458050/edit)
2. on the Subjects list (listing all Subjects in a Subject Set, i.e. `http://localhost:3000/projects/123/workflows/123/subject-sets/123/subjects/`), when a Subject is clicked

Question: why is it needed in two locations? Shouldn't it be enough that the Subject Viewer component handle the responsibility of fetching a Subject, when the component is loaded with a specific ID?

Counter argument: but it works.

Counter counter argument: but you're basically duplicating what's essentially an atomic function over two different components, so when you're testing "Is the fetch Subject function working?" you'll need to consider two scenarios: 1. when the user accesses the Subject Viewer/Editor page directly, and 2. when the user accesses the Subject Viewer/Editor by clicking on the Subjects list.

wgranger: That's a good point. I guess it would be more effective to only call `fetchSubject` when the editor is loaded with the given url param. Although it does work, I think you're right that this sort of code duplication is unnecessary.

wgranger:
Thanks! I've made a few changes based on the review:

- Remove the double `fetchSubject` call
- Show subject loading and error states
- Reset the subject zoom, pan states on index change

src/store/SubjectStore.js
shaunanoordin: Wait, goshdangit - you're completely right, Will. This specific project does NOT benefit from having a Subject queue. Please ignore my earlier comment, my brain was defaulting to "standard Zooniverse Classifier behaviour" while reading this code, instead of properly addressing the purpose of this project's Subject Store.

shaunanoordin:
Awesome! 👍 This PR looks as excellent as it can be with the latest changes, let's get it merged!

PR #47, opened 2019-11-01
Fetch Projects

wgranger:
This PR begins fetching projects from the TOVE backend. Currently, all projects are being returned from Tove as there are only a few in the database. Later, this will be changed to returning only those projects for the logged in user.

### DIFF
--- a/src/screens/ProjectsIndex/ProjectsPageContainer.js
+++ b/src/screens/ProjectsIndex/ProjectsPageContainer.js
@@ -13,8 +13,8 @@ function ProjectPageContainer () {
   if (store.auth.user && store.projects.asyncState === ASYNC_STATES.IDLE) {
     store.projects.getProjects()
   }
-  const ownerProjects = store.projects.ownerProjects
-  const collabProjects = store.projects.collabProjects
+  const ownerProjects = store.projects.ownerProjects || []
+  const collabProjects = store.projects.collabProjects || []
   const error = store.projects.error
 
   return (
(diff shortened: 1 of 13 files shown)


shaunanoordin:
Storybook is currently broken for this branch; a fix has been implemented in #50 . This branch requires rebasing to the new master, or (in my case) I'll just monkey-patch the corrected config for testing.

src/screens/ProjectsIndex/ProjectsPageContainer.js
```diff
@@ -12,30 +25,52 @@ export default function ProjectPageContainer () {
...
+        </Box>}
+
+      {store.projects.asyncState === ASYNC_STATES.LOADING &&
+        <Box justify='center' direction='row' gap='xsmall'>
+          <Text>Loading</Text>
+          <FontAwesomeIcon icon={faSpinner} spin />
+        </Box>}
+
+      {store.projects.asyncState === ASYNC_STATES.ERROR &&
+        <Text color='red' textAlign='center'>{error}</Text>}
```
shaunanoordin: Status indicators: nice 👍

src/screens/ProjectsIndex/ProjectsPageContainer.js
```diff
@@ -12,30 +25,52 @@ export default function ProjectPageContainer () {
...
+
+      {ownerProjects.length === 0 && collabProjects.length === 0 && store.projects.asyncState === ASYNC_STATES.READY &&
+        <Text textAlign='center'>
+          We couldn't find any transcription projects you participate in.
```
shaunanoordin: Edge cases: niiice 👍

wgranger:
Superagent vs. Frisbee

This was a suggestion from @srallen. I believe the idea was to try out Frisbee since it implements the [new fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). The monorepo may also start using Frisbee in the future if we have a good experience here.

PR #51, opened 2019-11-04
Remove Checkboxes from ResourcesTable

wgranger:
This PR removes checkboxes from the resources table as they're no longer needed [per InVision](https://projects.invisionapp.com/d/#/console/17925240/375544496/preview). I also updated the phrasing of the subject set download button above as that was the original impetus for the change.


wgranger:
Merging, as this isn't making huge changes aside from removing the checkbox

Issue #22, opened 2019-10-11
Allow Data Table Retrieval and Pagination


wgranger:
Allow data tables to paginate through data. Currently, there are data tables for projects, workflows, subject sets/groups, and subjects. Really, two questions exist here.

Display appropriate fields in the data table. Currently, each data table is hooked up with mock data. However, once the api is operating, the payload returned from the API will need to be turned into defined mobx models and fed into each data table.

- How do we know which data to return? eg. For user A, how does the API which *transcription* projects to return.
- From that project, we should be able to page through items. eg. Of the 225 subjects in workflow B, we should be able to page through 20 at a time.

wgranger:
The biggest lingering question here is how will we designate workflows and projects as being available in the text editor app? Will this be done through Panoptes or through the new text editor api?

wgranger:
This will be handled by the new text editor backend (TOVE). Each project will have to have an external effect set up in Caesar so TOVE can recognize these classifications.

PR #52, opened 2019-11-06
Fetch resources

wgranger:
Working towards #22. This PR fetches resources, but doesn't take care of paging quite yet. 

I am aware the site doesn't currently load resources from params (ex. hitting `/projects/123/workflows/456/subject-sets/789/subjects` will not load project 123). I'm not sure TOVE can return a resource by ID quite yet, so I'll have to wait until that functionality is in the backend. In the meantime, this PR should work by retrieving resources **if you begin on the projects page.** Also, some workflows don't have subject groups, so keep that in mind when selecting a project if you don't see any "subject sets" returning.


shaunanoordin:
EDIT: I just saw the comment:

> I am aware the site doesn't currently load resources from params

Ah, so this is what you meant about reading the comments! Changing my review to an Approve 👍

wgranger:
Thanks for the thorough review! Yes, this will need some more revisions once TOVE is ready to fetch resources by ID. As for the "back" button, I'm going to implement a version of this once I get the header to display the correct routes instead of always saying "Text Transcription Project."

PR #55, opened 2019-11-08
Add Dynamic Header Title

wgranger:
According to the [InVision](https://projects.invisionapp.com/d/#/projects/prototypes/17925240), the title should change depending on the current index page.


wgranger:
Thanks for the review! Subheaders now link back to their respective index page. Merging.

Issue #28, opened 2019-10-11
Display UI for Aggregated Lines


wgranger:
Display aggregation lines from Caesar as the transcription viewer will eventually allow for these aggregations to seemingly change in real time. The first step here is to get mock data from Caesar for a known subject like ASM and display those lines in the UI.

wgranger:
Closed via #85

PR #57, opened 2019-11-18
Show Transcription Lines

wgranger:
Alright, not quite the changes I had anticipated for this PR, but some important fixes here. My initial goal was to show transcription lines from Caesar on the `SubjectViewer`, but I spent most of my time reorganizing how the `SubjectViewer` works. Basically, the SV had no way of calculating scale based on the viewer and image size, so the transcription lines would look out of place when superimposed on top of the image. This is mainly because the transcription lines didn't have the correct scale and transform. 

That said, most of the changes here are reorganizing the subject viewer. However, the initial goal of displaying annotation dots and lines was achieved. The only caveat is that only _one_ transcription line is shown (as opposed to a line for each annotation on the [InVision](https://projects.invisionapp.com/d/#/console/17925240/374156988/preview)). I kept it at one line for two reasons: I still need to figure out how we're showing each transcription line, and I didn't want the PR to get too bloated.

Working towards #28


wgranger:
@shaunanoordin While staging is down, you'll have to comment out this line so you aren't redirected to the landing page for failure to login https://github.com/zooniverse/text-editor/blob/master/src/store/AuthStore.js#L15

Also, you'll probably have to head straight to the `/projects` path and go from there

shaunanoordin:
I've tried to get around our currently wonky Zooniverse Staging Authentication by disabling `AuthStore.js` line 15, but I get "null user" problems and `/projects` just lists nothing. 🤷‍♂  

Instead, I have another workaround - disabling Chrome's web security to allow CORS requests from the Staging API:

`> open -a Google\ Chrome --args --disable-web-security --user-data-dir`

Dangerous! But it wooorks

src/__snapshots__/Storyshots.test.js.snap
```diff
@@ -3766,7 +3766,7 @@ exports[`Storyshots SubjectViewer Default 1`] = `
             className="StyledBox-sc-13pk1d4-0 ekjAwv"
           >
             <span
-              className="StyledText-sc-1sadyjn-0 bmpSIR sc-kEYyzF eJZPaM"
+              className="StyledText-sc-1sadyjn-0 bmpSIR sc-eNQAEJ jtFbJk"
```
shaunanoordin: Ah, the most important change in this PR.

wgranger:
Good review. Thanks for picking out a few things I've overlooked. I prefer the `InteractionLayer` to move with the subject, because it seems odd a user would be able to move a subject if they aren't dragging on the subject. If the subject ever goes out of focus, the "reset" button can always be used on the image tools to put the image back in center. 

I'm not sure if users will need to interact with or click annotation lines, but swapping the order of the `InteractionLayer` and `AnnotationsPane` will let users interact with those lines if I should need to add click handlers later.

Good spot on `AsyncMessages` not actually working. It should be fixed now.

wgranger:
Important note! The test subject in this PR has blank space to the left and right of it (longer horizontal than vertical). I know it's weird, but keep that in mind when reviewing.

wgranger:
Merging this so I can begin rebasing the changes here into my store refactor

PR #65, opened 2019-12-11
Rewrite Mobx Stores to Use References

wgranger:
This PR does two things:

- Uses references and identifiers to store data in mobx resource stores
- Retrieves resources on initial load. This allows paths like `projects/123/workflows/123/groups` to load the correct resources on first load.
- Takes care of some of the "Subject Sets" vs "Groups" discrepancies 
- Add several missing component render tests to keep coverage from dropping too significantly after the refactor

Unfortunately, this PR is a bit bloated with the refactoring and restructuring of tests.


src/screens/Editor/Editor.js
```diff
@@ -25,11 +24,15 @@ function Editor ({ match }) {
   const store = React.useContext(AppContext)
   const editorBox = React.useRef(null)
 
-  const params = match.params
+  React.useEffect(() => {
```
shaunanoordin: So if I'm understanding this right, `React.useEffect()` is the equivalent of `componentDidUpdate` and `componentDidMount`.

I can see that this makes sense in the `componentDidMount` way - the resources are fetched as soon as the edit page [(example)](http://localhost:3000/projects/1764/workflows/11/groups/TEST1/subjects/51842/edit) is loaded.

What I'm curious about is that, shouldn't this also trigger when _observed values change?_ (And the Editor component is definitely being `observer()`-ed)

I can't seem to get this to happen though, so there seems to be no problem... I need to read up on `React.useEffect()` more

wgranger: Rerendering is dependent on which items are passed into the array at the end of the function, here being `[match, store]`. Or, you can pass in an empty array instead to say "only trigger once after render." 

However, since `match` and `store` are dependencies here, they are required to be passed into the array.

Check out the final yellow "note" section at the end of this page https://reactjs.org/docs/hooks-effect.html

src/screens/SubjectSetsIndex/SubjectSetPageContainer.js
```diff
@@ -12,21 +12,21 @@ function SubjectSetPageContainer({ history, match }) {
...
+  const groups = Array.from(store.groups.all.values())
 
   return (
     <Box margin='medium' fill='vertical'>
       <ResourcesTable
         columns={columns}
-        data={store.groups.all}
-        resource='Subject Sets'
+        data={groups}
+        resource='groups'
```
shaunanoordin: It seems that most of the text labels and class names that use "Subject Set" (or "Set") have been renamed to "Group", which is cool.

The following still seem to be called "Subject Set" though:
- src/screens/SubjectSetsIndex
- src/screens/SubjectSetsIndex/SubjectSetPageContainer
- blablabla/HeaderButton/DownloadSetDataContainer

wgranger: Good point. Fixed.

src/store/AppStore.js
```diff
@@ -19,11 +19,40 @@ const AppStore = types.model('AppStore', {
...
   subjects: types.optional(SubjectStore, () => SubjectStore.create({})),
-  initialised: types.optional(types.boolean, false),
+  initialized: types.optional(types.boolean, false),
   projects: types.optional(ProjectsStore, () => ProjectsStore.create({})),
   modal: types.optional(ModalStore, () => ModalStore.create({})),
   transcriptions: types.optional(TranscriptionsStore, () => TranscriptionsStore.create({})),
   workflows: types.optional(WorkflowsStore, () => WorkflowsStore.create({})),
-})
+}).actions(self => ({
+  getResources: flow (function * getResources(params) {
```
shaunanoordin: Noice, a super function that fetches resources in order of dependencies. I like me a super function.

However, I haven't had the chance to really try to break this function. e.g. is it possible to get `self.projects.reset()` to reset _without_ resetting the dependent workflows? Will there be a problem if I chuck in `project = 0`, which will evaluate to `false` in `!!params.project` ? etc etc etc

But as far as _normal functionality_ looks though, this is a good ol' 👍

src/screens/Editor/Editor.js
shaunanoordin: 👍

src/store/WorkflowsStore.js
```diff
@@ -30,27 +39,59 @@ const WorkflowsStore = types.model('WorkflowsStore', {
     }
   }),
 
-  selectWorkflow: function(workflow) {
-    if (!workflow) {
-      getRoot(self).groups.setGroups([])
-      self.current = Workflow.create()
-      return
+  getWorkflow: flow (function * getWorkflow(id) {
+    console.log('LOOK');
```
shaunanoordin: Seen

src/store/WorkflowsStore.js
```diff
@@ -30,27 +39,59 @@ const WorkflowsStore = types.model('WorkflowsStore', {
...
+    if (!id) return undefined
+    console.log('TWICE');
```
shaunanoordin: Seen again

wgranger: Whoops!

PR #68, opened 2019-12-13
Index pagination

### LINKED ISSUE #22: Allow Data Table Retrieval and Pagination

wgranger:
This PR sets up pagination in all of the resource index pages. Note: the `GroupsStore` is done a bit different than the other resource stores.

Closes #22


shaunanoordin:
## PR Review...?

This PR should, in theory, allow resources pages - like the Workflows page [(example)](http://localhost:3000/projects/1764/workflows) and Subjects page [(example)](http://localhost:3000/projects/1764/workflows/11/groups/TEST1/subjects) - to have pagination.

Code looks legit, but I can't seem to test the feature because I haven't found any resource with more than 4 items in 'em. 🤔 

### Status

🤷‍♂

wgranger:
That's a good point. I had to hack the code a bit to get pagination to appear, but that's unnecessary for you to do. Let me see if we can get some more subjects attached to a group so you can test this out accurately. 

Sorry I forgot to get that squared away before posting this PR ☹️

PR #70, opened 2019-12-16
Open transcription panel

wgranger:
This PR opens the transcription pane when clicking on a transcription to the right. Check out the [InVision](https://projects.invisionapp.com/d/#/console/17925240/371813757/preview). I had to move around some of the elements padding and margins so this would appear to "overflow" the parent.

**Note**: Most of the line changes here are updating the snapshots.

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -11,1950 +11,2066 @@ exports[`Storyshots AggregatedTranscriptions Default 1`] = `
       className="StyledBox-sc-13pk1d4-0 bdFdoP"
     >
       <div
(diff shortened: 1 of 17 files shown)


wgranger:
Good review. I've moved around a few items and updated the tests. However, I wasn't able to get the transcription panel to toggle open when moving the box order. That would not be the intended behavior.

shaunanoordin:
## PR Review (Update)

Yo @wgranger I was just checking on this PR and it's looking great.

- Latest changes makes the UI look good
- Re: _"However, I wasn't able to get the transcription panel to toggle open when moving the box order."_ this issue seems to be solved in the latest commits. It was extant in ef8a522 but gone by a65e911 - maybe it was solved by the new `stopEvents()` code?

In any case, this PR's looking good and should be merged soon 👍

PR #71, opened 2019-12-17
Search Modal Functions

### LINKED ISSUE #21: Search Modal Functionality
Hook up the search modal with functionality to search by Zooniverse ID or External ID. The modal should also be able to search by several filters including:  

**Approval Status:**
- Unreviewed
- In progress
- Ready for Review
- Approved

**Additional Filters**
- Flagged
- Low consensus score

wgranger:
This is WIP because we're still waiting on a couple items before this is ready.

Closes #21 

1) Waiting for #68 to merge because I'll have to keep track of the current query when paginating. What if somebody has an active search filter? They'll want to retain that filter when paginating.
2) There are several items that aren't yet filterable on the Tove end (`internal_id`, `low_consensus_score`, filtering by status string instead of enum integer).

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -3134,7 +3134,7 @@ exports[`Storyshots SearchModalContainer Default 1`] = `
       className="StyledBox-sc-13pk1d4-0 jLJOXa"
     >
       <div
-        className="StyledBox-sc-13pk1d4-0 juVmUb"
+        className="StyledBox-sc-13pk1d4-0 kULYLq"
       >
         <div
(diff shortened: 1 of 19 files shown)


wgranger:
This is now ready for review. The only change is that "Low Consensus" was removed from the design in lieu of sorting columns, which will be implemented in a future PR. 

Design link -> https://projects.invisionapp.com/d/#/console/17925240/378854810/preview

wgranger:
Searching by `internal_id` is not yet supported by the TOVE backend. There's an [open issue](https://github.com/zooniverse/tove/issues/72) for it.

I can add some constants quickly. The idea of having a "reset search" button wasn't in the design, so if we add this it'll likely be in a future PR.

wgranger:
Added constants to this PR. Merging.

PR #79, opened 2020-01-24
Update Index Page Items

wgranger:
With the backend being fleshed out, more items are available on TOVE resources to further populate the index pages.


wgranger:
Merging, as these are small updates to the index page based on recent TOVE changes

PR #85, opened 2020-01-31
Show Extract Lines

wgranger:
There's a lot of functionality going on here to get extract lines mapped to reduction lines. However, this PR should do the heavy lifting there. 

One item of note, this PR will only get extracts working for ASM on staging. There's a bit of work we still need to do to backfill existing extracts to use a pre-determined extract key that will work with ALICE. Note here that we use `ext-17`, the one used for ASM.

For now, this PR does a couple new things:
- Show extract lines from previous users
- Correctly colors extract and reduction lines
- Allows toggling line visibility on and off


wgranger:
A big change here is how we display svg lines. This PR only displays two points (`x1, x2, y1, y2`), whereas before we would iterate through all points in a line.

The main question is, will the text editor be used by tasks with > 2 points? If so, that will have to change in this PR. However, this _should_ work if all tasks have only a beginning and end line.

wgranger:
PR should be good to go. Instead of relying on an annotation always having two points, we now just get the last point as the final point.

src/components/SubjectViewer/components/AnnotationsPane/AnnotationsPaneContainer.js
```diff
@@ -2,19 +2,42 @@ import React from 'react'
 import { observer } from 'mobx-react'
 import AppContext from 'store'
 import AnnotationsPane from './AnnotationsPane'
-import { constructCoordinates } from 'helpers/parseTranscriptionData'
+import { constructCoordinates, constructText, mapExtractsToReductions } from 'helpers/parseTranscriptionData'
 
 function AnnotationsPaneContainer({ x, y }) {
-  let parsedLines = []
+  let [ extractLines, reductionLines ] = [[], []]
```
shaunanoordin: Is this just a fancy way of writing `let extractLines = [], reductionLines = []` ?

src/helpers/parseTranscriptionData.js
```diff
@@ -1,8 +1,13 @@
 function constructCoordinates(line) {
   const points = [];
+
   if (line && line.clusters_x && line.clusters_y) {
-    line.clusters_x.forEach((point, i) => {
-      points.push({ x: line.clusters_x[i], y: line.clusters_y[i] });
+    const lastPoint = line.clusters_x.length - 1
```
shaunanoordin: I assume that since you're trusting the results from Tove, there _shouldn't_ be any conditions where...

1. line.clusters_x and line.clusters_y have different lengths
1. line.clusters_x will have at least (exactly?) 2 items

src/components/SubjectViewer/components/AnnotationsPane/SVGLines.js
```diff
@@ -0,0 +1,72 @@
...
+
+export default function SVGLines({ lines, isExtract, reductionIndex }) {
+  const circleWidth = isExtract ? 4 : 10
+  const dashArray = isExtract ? '4' : '0'
+  const strokeWidth = isExtract ? '0.5' : '3'
+
+  return (
+    <g>
+      {lines.map((line, index) => {
+        const color = indexToColor(reductionIndex % 12)
```
shaunanoordin: Oh wait, I forgot to comment on this. Some thoughts: the modulo `%` action should happen _inside_ the `indexToColor()` function itself. This would be a good idea if, e.g. you later modify `indexToColor()` to  have more/less than 12 colours, so you only need to modify the code in one file - `indexToColor.js`

src/helpers/parseTranscriptionData.js
wgranger: Correct. There should always be the same amount of X and Y points from TOVE, since each annotation point will have an X and Y coordinate. Every project using TOVE/ALICE should only have two points, but using `length - 1` will still allow lines to render with the first and last points.

src/components/SubjectViewer/components/AnnotationsPane/SVGLines.js
wgranger: Ah, good point. I'll make that change quickly.

PR #86, opened 2020-01-31
Avoid Two hour timeout

### LINKED ISSUE #82: Handle Two Hour Timeout
Some thinking needs to go into this, because the Frisbee library we are using doesn't have the luxury of a `beforeEveryRequest` function like what we have on the panoptes client. We need to find a workaround in the case someone is idle for two hours, loses their bearer token, and picks up the app again. The app would likely crash because the bearer token has expired then, but it needs to be refreshed.

wgranger:
As we know from ASM, Panoptes tokens will expire after two hours. The difference between this app and ASM is that we're using auth here, where we used oauth in ASM. Oauth needed a redirect to allow a bearer token refresh. Auth does not require a redirect, so we can refresh the token staying in the same spot.

However, all calls need to be wrapped in a bearer token check to make sure we always have a valid token and are sending that back to TOVE. This PR should handle all of that. However, depending on when this merges, I'll have to update other open PRs to use this sort of wrapper around outgoing TOVE calls.

Closes #82

### DIFF
--- a/src/store/ClientStore.js
+++ b/src/store/ClientStore.js
@@ -73,9 +73,12 @@ const ClientStore = types.model('ClientStore', {
   setBearerToken: (token) => {
     // If Token is null, we need to remove jwt tokens
     // If token is a string with a length, we need to set tokens
-    if (token === null || token.length) {
+    if (token) {
       self.tove.jwt(token)
       self.toveZip.jwt(token)
+    } else {
+      self.tove.jwt()
+      self.toveZip.jwt()
     }
   }
 }))
--- a/src/store/ClientStore.js
+++ b/src/store/ClientStore.js
@@ -64,10 +64,10 @@ const ClientStore = types.model('ClientStore', {
     })
   }),
 
-  patch: flow(function* get (request) {
+  patch: flow(function* get (request, body) {
     return yield auth.checkBearerToken().then((token) => {
       self.setBearerToken(token)
-      return self.tove.patch(request)
+      return self.tove.patch(request, body)
     })
   }),
 
--- a/src/store/ClientStore.js
+++ b/src/store/ClientStore.js
@@ -64,6 +64,13 @@ const ClientStore = types.model('ClientStore', {
     })
   }),
 
+  patch: flow(function* get (request) {
+    return yield auth.checkBearerToken().then((token) => {
+      self.setBearerToken(token)
+      return self.tove.patch(request)
+    })
+  }),
+
   setBearerToken: (token) => {
     if (token !== self.bearerToken) {
       self.bearerToken = token
--- a/src/store/ClientStore.js
+++ b/src/store/ClientStore.js
@@ -6,7 +6,6 @@ import auth from 'panoptes-client/lib/auth';
 
 const ClientStore = types.model('ClientStore', {
   aggregator: types.optional(types.frozen({}), null),
-  bearerToken: types.optional(types.string, ''),
   tove: types.optional(types.frozen({}), null),
   toveZip: types.optional(types.frozen({}), null)
 }).actions(self => ({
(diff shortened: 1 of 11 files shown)


wgranger:
Marking "do not merge" as this will have to be updated to wrap all api calls once all current PRs are merged.

wgranger:
This should be good to go now. I needed to update the branch to wrap `.patch` requests to Tove, but I think this should handle all requests. I don't believe any of the other requests to external services (Panoptes, Caesar) require auth and necessitate an updated bearer token.

src/store/TranscriptionsStore.js
```diff
@@ -91,7 +91,7 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
   }
 
   const checkIfLocked = flow(function * checkIfLocked() {
-    const client = getRoot(self).client.tove
+    const client = getRoot(self).client
```
srallen, line 94: You can use a destructured syntax here and the other similar lines in this same file: `const { client } = getRoot(self)`

src/store/WorkflowsStore.js
```diff
@@ -31,7 +31,7 @@ const WorkflowsStore = types.model('WorkflowsStore', {
   fetchWorkflows: flow (function * fetchWorkflows(id, page = 0) {
     self.page = page
     self.asyncState = ASYNC_STATES.LOADING
-    const client = getRoot(self).client.tove
+    const client = getRoot(self).client
```
srallen, line 34: Same comment about destructuring syntax.

wgranger:
I decided against using panoptes-js because there is still a good bit of TODOs in that package and it's not quite capable of what I'm doing here. Panoptes-js currently cannot fetch project roles, subjects, or users, and I believe it can only fetch a single project by a string instead of an array of projects with a single call.

I decided to store the bearer token in the store to avoid setting the bearer and jwt tokens every time a call is made to TOVE. It seemed more beneficial to optimize the store in this way. See the `setBearerToken` function in the `ClientStore`, which conditionally checks that the token has changed before setting the token on the store and running a couple functions on the two Frisbee objects.

srallen:
If it's helpful, here's the internal auth method that is called when the token is expired: https://github.com/zooniverse/panoptes-javascript-client/blob/93a3f2104855b15cea44134d2b01fa6fcb2d0554/lib/auth.js#L63-L90

wgranger:
Nice. Thanks for the references. It makes sense that storing tokens in the store is probably a bad idea. 

I've removed any token references from the store. It was only being used to avoid setting the jwt tokens on the client with each call. However, I don't think it is a huge issue to make that call, as there isn't much to the Frisbee [function itself](https://github.com/niftylettuce/frisbee/blob/181867c796bcfd88316850f366df520d8c134f77/src/index.js#L567).

Now, JWT tokens will only be set if the token is `null` (eg. logout) or if it is a string with a truthy length.

src/store/ClientStore.js
```diff
@@ -56,9 +56,27 @@ const ClientStore = types.model('ClientStore', {
...
+
   setBearerToken: (token) => {
-    self.tove.jwt(token)
-    self.toveZip.jwt(token)
+    // If Token is null, we need to remove jwt tokens
+    // If token is a string with a length, we need to set tokens
+    if (token === null || token.length) {
+      self.tove.jwt(token)
+      self.toveZip.jwt(token)
+    }
```
srallen, line 79: One comment based on the frisbee docs:

> api.jwt(token) - helper function that sets a JWT Bearer header. It accepts the jwt_token as a single string argument. If you simply invoke the function null as the argument for your token, it will remove JWT headers.

If the token isn't defined, perhaps this should call `self.tove.jwt()` and `self.toveZip.jwt()`? I'm not sure if there is any other handling this app should have, but the header likely should be removed if it had been set.

wgranger:
Oops, missed all of the destructuring with the previous commits. I had to mock out the JWT function on each test using a Frisbee client. If I were to leave it in, the calls would error out since the Frisbee client would think I were unauthorized.

PR #88, opened 2020-02-04
Download Approved Data by Group

wgranger:
This PR allows a modal to appear when clicking to download all approved subjects. Beforehand, the modal was only hooked up for a single subject, but this reuses the component while making a couple design changes when downloading approved subjects on a group.


wgranger:
Ahhh, good point. I need to disable the button of none of the subjects are approved.

wgranger:
Made a slight tweak to disable the button if no transcriptions are approved.

PR #78, opened 2020-01-23
Show Transcription Text

wgranger:
This PR allows transcriptions to be selected from the panel to the right of the subject viewer. From there, users are able to do a few things:
- Select a transcription
- Replace the selected transcription with the toggled option
- If not saving the original algorithm option, the new selection should be saved as `edited_consensus_text` and the editing user should be saved as `line_editor`
- Type in and select a different transcription
- Add "flagged" or "seen" markers to a line (choosing the green dot or red flag)
- Save the current transcription state

You can check out the design on the [InVision](https://projects.invisionapp.com/d/#/console/17925240/371813757/preview). Note, there are a few design elements that aren't yet completed (user name and date under a transcription, add and delete functionality). Those will be added in a future PR.

Closes #30 
Closes #75 
Working towards #33

PR #89, opened 2020-02-10
Transcription Pane Improvements

wgranger:
This PR covers many of the UI issues found in the review of #78. However, I'm holding off on the issue of the checkbox being too small for a later UI audit, as that is worth a longer discussion.


wgranger:
Functionality still needs to be added for those two buttons. They should be working within the next couple weeks.

Issue #33, opened 2019-10-11
Line Viewer Functions


wgranger:
Add functionality to line viewer. This modal has the most amount of functions in the app. Users should be able to add line below, delete line, replace with selected line, and write a new line. 

Users are also able to both flag a line and mark it as seen.

PR #91, opened 2020-02-11
Delete Transcription Lines

wgranger:
This PR handles deleting an annotation line and saves the transcription on deletion. Be careful with this, as it is currently irreversible. I'd like to get #89 merged first so I can clean up some code in the `LineViewerContainer` to get testing up. 

[Invision Design](https://projects.invisionapp.com/share/KBVXAK04S2C#/screens)

Working towards #33

### DIFF
--- a/src/store/TranscriptionsStore.js
+++ b/src/store/TranscriptionsStore.js
@@ -59,11 +59,13 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
   },
 
   deleteCurrentLine: function deleteCurrentLine() {
-    const index = getRoot(self).subjects.index
-    const page = self.current.text.get(`frame${index}`)
-    page.splice(self.activeTranscriptionIndex, 1)
-    self.saveTranscription()
-    self.setActiveTranscription()
+    if (Number.isInteger(self.activeTranscriptionIndex)) {
+      const index = getRoot(self).subjects.index
+      const page = self.current.text.get(`frame${index}`)
+      page.splice(self.activeTranscriptionIndex, 1)
+      self.saveTranscription()
+      self.setActiveTranscription()
+    }
   },
 
   retrieveTranscriptions: flow(function * retrieveTranscriptions(query) {
--- a/src/store/TranscriptionsStore.js
+++ b/src/store/TranscriptionsStore.js
@@ -58,6 +58,14 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     })
   },
 
+  deleteCurrentLine: function deleteCurrentLine() {
+    const index = getRoot(self).subjects.index
+    const page = self.current.text.get(`frame${index}`)
+    page.splice(self.activeTranscriptionIndex, 1)
+    self.saveTranscription()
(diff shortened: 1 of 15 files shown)


src/store/TranscriptionsStore.js
```diff
@@ -56,6 +56,14 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     })
   },
 
+  deleteCurrentLine: function deleteCurrentLine() {
+    const index = getRoot(self).subjects.index
+    const page = self.current.text.get(`frame${index}`)
+    page.splice(self.activeTranscriptionIndex, 1)
+    self.saveTranscription()
+    self.setActiveTranscription()
+  },
```
shaunanoordin: This function looks legit, and basic tests shows that it works fine... (e.g. `self.setActiveTranscription(nothing)` correctly resets the actively selected item after it's been deleted, which was the first thing I looked for) 

I'm wondering if there's wisdom in adding a check to make sure `(self.activeTranscriptionIndex !== null && self.activeTranscriptionIndex !== undefined)`, or `(Number.isInteger(self.activeTranscriptionIndex))` as a safety. I mean, the _UI_ makes it practically impossible for `activeTranscriptionIndex` to be undefined when the Delete modal appears, so I might just be too paranoid.

(Note: `arr.splice(undefined, 1)` works the same as `arr.splice(0, 1)`)

(Reminder: please don't use `(!!self.activeTranscriptionIndex)` since that'll trip up when the index is 0.)

shaunanoordin:
## Testing Fun Times

Quirks unrelated to this PR, but came up while I was testing, so they might be worth noting in separate issues.

1. if a Delete attempt is performed when the network is suddenly offline, ⚠️ there is no notice that the call-to-API action failed.
  - ⚠️ Bigger Question: how many other call-to-API actions (e.g. rearranging lines, saving changes to selected transcription line, changing flagged status, etc) don't give warnings to the user if they fail?
  - Has this scenario been planned for yet?
2. It's POSSIBLE to change the "selected transcription index" to something invalid.
  - For example, a Subject with 2 pages - 7 lines on first page, 5 lines on the second page.
  - If I select line number 7 on page 1, the Subject Editor will show me details for that line of aggregated text 👌 
  - ⚠️ If I **change the page** to the second page, them the Subject Editor will attempt to **show the details for the non-existent 7th line on the second page.**
  - One good thing is that the _app doesn't crash immediately_ upon receiving a nonsense index-out-of-bounds error, so credit where it's due!

wgranger:
Aside from some rebasing, it looks like I need to make a few changes to this branch. 

The transcription and page index mismatch is definitely something I should address before this merges. 

Aside from that, there is an issue (#96) that will cover displaying errors when saving, but error handling as a whole is probably something we should put a bit more thought into for this ap.

src/store/TranscriptionsStore.js
wgranger: Although a user _shouldn't_ be able to delete a transcription without one selected, I think it is a smart idea to add a check here.

PR #92, opened 2020-02-13
Undo manager

### LINKED ISSUE #29: Undo Button Functionality
Provide functionality for the undo button. This is taking advantage of the snapshot feature in mobs which will allow the state tree to go back in time to undo past changes.

wgranger:
Closes #29 

This PR hooks up the undo functionality to the text editor. Using some [mobx middleware](https://github.com/mobxjs/mobx-state-tree/tree/master/packages/mst-middlewares#undomanager) we can now go back to previous patches. 

The `UndoManager` is only linked to the `TranscriptionsStore`, with some patches omitted from the manager. Because of this, I had to bring the `changeIndex` action over to the `TranscriptionsStore` from the `SubjectsStore`.


wgranger:
Marking WIP to auto save the transcription with each undo.

shaunanoordin:
## PR Review (Update)

Well, at least I've discovered what's causing my sporadic _"Undo button is still clickable after navigating away and returning to Subject"_ errors:

- if an 'Edit Transcription' popup is open, and I navigate away from the Subject editor page (to the Subjects listing page) without closing it...
- ...then when I return to the same Subject editor page, the Undo stack keeps at least 'one level of action' (i.e. you can click Undo, and it'll open the Edit Transcription popup)

Screen Shot 2020-03-03 at 18 27 38
_Step 1: open an Edit Transcription popup. Step 2: navigate back to the Subjects listing page, by clicking on the Subjects listing page breadcrumb ('12345') on the upper left_

Screen Shot 2020-03-03 at 18 27 46
_Step 3: go back to the same Subject to edit it._

Screen Shot 2020-03-03 at 18 27 56
_Step 4: note that the Undo button is now clickable._

Notes:

- This 'persistent Undo' bug is only valid for one action (i.e. re-opening the popup.)
- ANY other (tested) action doesn't result in the persistent Undo bug - e.g. if the last action (before navigating away to the Subject listing page) was to rearrange lines, then there's no Undo button when the Edit Subject page is revisited.

Unknown/Not yet tested:

- What happens if the list of Subjects has more than 1 Subject I can navigate to?
  - If I go to Subject A, open an 'edit transcription' popup (and not close it), navigate back to the Subjects list, then navigate to Subject B...
  - ...will the 'persistent Undo' bug still be there... but for the wrong Subject?

I'm marking this as an _eeehhh_ 🤷‍♂ level bug. I mean, it's not particularly destructive, and you shouldn't spend dozens of hours to solve a bug that only 1% of editors might ever notice and has little chance of triggering catastrophic errors.

But it does highlight the complexity of the Undo feature.

wgranger:
I'd like to go through this PR once again before merging to take care of some of the items you mentioned. I can think of a couple reasons for this odd behavior.

wgranger:
I think most of the issues you were experiencing are taken care of with the most recent commit. The issue was that the `undoManager.clear()` was occurring _before_ some changes were made to the store with the mount/useEffect. Because of that, some changes from `setActiveTranscription()` and possibly unexpected changes with `setResources()` were added to the undo history after it was cleared.

Instead, what I want to do is finish mounting the page with all of the changes, and finally clear the undoManager. That will allow the user to see the page fresh without a state to revert back to.

PR #100, opened 2020-02-18
Deploy App Using React Scripts

### LINKED ISSUE #94: Deploy App
The app needs to be published using S3. The biggest change here will be getting the deployment script situated in the `package.json`

wgranger:
I'm utilizing `react-scripts` here to handling building the app, then appending the correct path `alice.zooniverse.org` for deployment and `preview.zooniverse.org/alice` (alice.preview.zooniverse.org) for staging to send the site to the S3 bucket with `publisssh`. 

I've also created a `config.js` file to handle URLs that must dynamically change depending on the environment.

Closes #94

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "scripts": {
     "_publish": "publisssh ./build \"zooniverse-static/$PATH_ROOT\"",
     "deploy": "export PATH_ROOT=alice.zooniverse.org; yarn build && yarn _publish",
-    "stage": "export PATH_ROOT=preview.zooniverse.org/alice; yarn build && yarn _publish",
+    "stage": "export REACT_APP_ENV=development; export PATH_ROOT=preview.zooniverse.org/alice; yarn build && yarn _publish",
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test --coverage --watchAll=false --silent",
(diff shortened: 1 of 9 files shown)


wgranger:
This can be tested with the dry run flag `-d` https://www.npmjs.com/package/publisssh

wgranger:
This should be working, but I had to hack the `panoptes-client` in the `node_modules` to get [this fix](https://github.com/zooniverse/panoptes-javascript-client/pull/128) in

wgranger:
Also removed the `@zooniverse/panoptes-js` dependency so I won't have to do a fix similar to the PJC one on that package. It was only being used to fetch subjects.

wgranger:
This should be working with the newly updates panoptes-client

package.json
```diff
@@ -8,7 +8,6 @@
     "@fortawesome/react-fontawesome": "^0.1.4",
     "@storybook/addon-storyshots": "^5.2.5",
     "@zooniverse/grommet-theme": "^2.1.0",
-    "@zooniverse/panoptes-js": "^0.1.1",
```
mcbouslog: I'm not super familiar with `panoptes-js`, was there anything unique it was doing, that PJC doesn't do?

mcbouslog:
Oh can be in subsequent PR - add urls (staging and production) to README

package.json
mcbouslog: Related updated requests look good, nevermind.

Issue #84, opened 2020-01-28
Handle More than 50 Project Roles


wgranger:
Currently, the `ProjectsStore` requests 50 roles for a user, and maps those roles to a project. 
https://github.com/zooniverse/text-editor/blob/master/src/store/ProjectsStore.js#L31

However, what will happen when a user has roles on more than 50 projects? This is an unlikely scenario, but the situation could occur with some admin.

wgranger:
Closed via #101

PR #101, opened 2020-02-18
Adjust Role Titles to Use Admin and Editor

### LINKED ISSUE #98: Adjust Role Titles
The original roles (Researcher, Volunteer, and Viewer) should change to better comply with existing Panoptes roles. The new roles will be:

Admin - full access
Editor - full access sans approval
Viewer - minimal access

wgranger:
Closes #98 

Mainly just a change in terminology. This is to avoid the former confusion where a user with a PFE role of "researcher" was not given the "researcher" role in ALICE.

### DIFF
--- a/src/store/ProjectsStore.js
+++ b/src/store/ProjectsStore.js
@@ -9,7 +9,7 @@ const ROLES = {
 }
 
 const adminRoles = ['owner', 'collaborator']
-const editorRoles = ['expert', 'researcher', 'moderator']
+const editorRoles = ['expert', 'researcher', 'moderator', 'scientist']
 
 const Project = types
   .model('Project', {
--- a/src/store/ProjectsStore.js
+++ b/src/store/ProjectsStore.js
@@ -55,9 +55,7 @@ const ProjectsStore = types.model('ProjectsStore', {
       const resources = JSON.parse(response.body)
       const ids = resources.data.map(project => project.id)
       const projects = yield apiClient.type('projects').get({ id: ids.toString(), cards: true })
-      console.log('get the roles');
       yield self.getRoles(ids)
-      console.log('DId we get em');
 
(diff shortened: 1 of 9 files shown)


wgranger:
I think I need to optimize it a bit to close #84. If you are associated with more than 50 projects, it might be that your role for one of the text editor projects isn't being retrieved.

I'm going to try to get a collection of project roles based on projects in the text editor instead of _all_ projects associated with a user.

wgranger:
Closes #84. Alright, give this another go. I think this should fix the issue, but let me know if not. Either way, this closes another issue linked.

src/store/ProjectsStore.js
```diff
@@ -3,13 +3,13 @@ import apiClient from 'panoptes-client/lib/api-client.js'
...
-  VOLUNTEER: 'Volunteer',
+  ADMIN: 'Admin',
+  EDITOR: 'Editor',
   VIEWER: 'Viewer'
 }
 
-const researcherRoles = ['owner', 'collaborator']
-const volunteerRoles = ['expert', 'researcher', 'moderator']
+const adminRoles = ['owner', 'collaborator']
+const editorRoles = ['expert', 'researcher', 'moderator']
```
mcbouslog: Oh I think I see why it's still happening! I think `'scientist'` needs to be added to this array, per 
Screen Shot 2020-02-26 at 10 35 16 AM
Oy the roles are a quagmire! But I think if this should do it.

wgranger: Panoptes defines the "researcher" role as a "scientist?" That's very confusing. I'll make the edit and ask you to give it one more spin to make sure all is sorted out. Thanks for your tedious review.

wgranger:
Good eye on the review. This should be good to go.

PR #103, opened 2020-02-25
Save status

### LINKED ISSUE #96: Show Saving Status
In lieu of the "save" button, we are now showing a save status. When a transcription fails to save, the error should appear as a banner. https://projects.invisionapp.com/d/main#/console/17925240/405495485/preview

wgranger:
Closes #96 

This PR removes the save button in lieu of a save status icon which will turn red and show an error when having trouble connecting to the backend. Note, the fetch api will not go straight to a `catch` in the case of a 404 or similar error, but the fetch api will go straight to the catch in the event of no internet (since a backend cannot provide an error code). More info on [this github discussion](https://github.com/github/fetch/issues/201#issuecomment-308213104).

Because of this, I try to handle errors the following way:
- 400 error: Display the error from the backend and ask the user to refresh
- 500 error: Display something unexpected happened and ask the user to refresh
- Loss of connection ("Failed to fetch"): Display internet is out, ask user to stay put as mobx should be saving updates as they go
- Default: Display unknown error and ask the user to refresh

[Error mockup](https://projects.invisionapp.com/d/#/console/17925240/405495485/preview)
[Changes saved mockup](https://projects.invisionapp.com/d/#/console/17925240/379453982/preview)


wgranger:
I'm tempted to address some of these issues on this branch. However, keeping MVP in mind, I'm going to open an issue outlying the main points here with a link to the PR. I intend to return to the UI thoughts after we reach MVP.

PR #104, opened 2020-02-27
Allow Transcriptions Index Page to Load Without Text Object

wgranger:
Due to a recent change in [how we serialize the text object](https://github.com/zooniverse/tove/pull/135), the index page was no longer loading without text objects. This PR should fix that issue.

A later backend change is coming which will allow items once held in the text object (consensus score, transcribed lines, and pages) to be pulled from the transcription object.


wgranger:
Merging, as this fix is needed to get Alice working

PR #105, opened 2020-02-28
Refactor and Save Original Transcriber

### LINKED ISSUE #90: Save the Original Transcriber
When selecting a new transcription from several options, we must save the `original_transcriber`of the line. This is the user who originally transcribed the line.

wgranger:
Closes #90 

In attempting to save the original transcriber, I had to do more mapping of extracts and reductions. In doing so, I found many opportunities for refactoring. This PR refactors many of those mapping functions while also rewriting the `TranscriptionsStore` tests as they were getting a bit unwieldy and needed to be straightened up some.

### DIFF
--- a/src/helpers/parseTranscriptionData.js
+++ b/src/helpers/parseTranscriptionData.js
@@ -1,3 +1,5 @@
+const SLOPE_BUFFER = 10;
+
 function constructCoordinates(line) {
   const points = [];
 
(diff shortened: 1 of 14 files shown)


wgranger:
WIP as this will take a while to rebase with recent changes

mcbouslog:
From http://localhost:3000/projects/1764/workflows/3085/groups/12345/subjects/85070/edit, if I click a line to review, I get hook related error, from LineViewerContainer:
Screen Shot 2020-03-09 at 3 35 14 PM

Maybe the conditional https://github.com/zooniverse/text-editor/pull/105/files#diff-66741304c007c4f5019aa447d3138c18R43 is causing the issue?

src/helpers/parseTranscriptionData.js
```diff
@@ -32,28 +34,50 @@ function mapExtractsToReductions(
   reductionIndex = 0,
   reductionText = [],
   subjectIndex = 0,
+  extractUsers = {}
 ) {
   const result = []
+  reduction.user_ids && reduction.user_ids.forEach((userId, idIndex) => {
+    // lets find that user's annotation
+    const userClassificationsOnSubject = extractsByUser[userId]
+    for (let i = 0; i < userClassificationsOnSubject.length; i++) {
```
mcbouslog: Getting TypeError warning `Cannot read property 'length' of undefined at parseTranscriptionData.js:43` I think related to if this is undefined.

wgranger:
This should be good to go now. Extracts weren't being loaded because I was still using the old `ext17` key instead of the new, more general `alice` key.

wgranger:
There are only three subjects in the current subject set at https://antislaverymanuscripts.preview.zooniverse.org/ and all subjects are currently in TOVE. New classification data will not be added to the transcription once it is in the tove DB.

PR #107, opened 2020-03-02
Add Line Function

### LINKED ISSUE #33: Line Viewer Functions

wgranger:
This PR gives functionality to the "Add Line" buttons both when a transcription is selected and on the main editor page (see [InVision](https://projects.invisionapp.com/d/#/console/17925240/371813757/preview)). 

When [adding a new transcription](https://projects.invisionapp.com/d/#/console/17925240/401676674/preview), when a user does not input text: 
- Closing the panel should delete the transcription
- Adding a line should be disabled
- The "seen" and "flagged" flags should be disabled

After inputting text, the new line will behave as the other lines.

This PR will close #33

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "stage": "export REACT_APP_ENV=development; export PATH_ROOT=preview.zooniverse.org/alice; yarn build && yarn _publish",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test --coverage --watchAll=false --silent",
+    "test": "react-scripts test --coverage --watchAll=false",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "stage": "export REACT_APP_ENV=development; export PATH_ROOT=preview.zooniverse.org/alice; yarn build && yarn _publish",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test --coverage --watchAll=false",
+    "test": "react-scripts test --coverage --watchAll=false --silent",
     "eject": "react-scripts eject",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -11,7 +11,7 @@ exports[`Storyshots AggregatedTranscriptions Default 1`] = `
       className="StyledBox-sc-13pk1d4-0 bdFdoP"
     >
       <div
-        className="StyledBox-sc-13pk1d4-0 dPQuSd sc-jzJRlG byHcWR"
+        className="StyledBox-sc-13pk1d4-0 dPQuSd sc-fjdhpX fCEyoV"
       >
         <div
           className="StyledBox-sc-13pk1d4-0 ikTqht"
@@ -25,15 +25,25 @@ exports[`Storyshots AggregatedTranscriptions Default 1`] = `
             >
               Transcribed Text
             </span>
-            <span
-              className="StyledText-sc-1sadyjn-0 fNQaFh sc-fjdhpX jMSAXb"
-              size="xsmall"
+            <button
+              className="StyledButton-sc-323bzc-0 jfbKsv"
+              onBlur={[Function]}
+              onClick={[Function]}
+              onFocus={[Function]}
+              onMouseOut={[Function]}
+              onMouseOver={[Function]}
+              type="button"
             >
-              Add Line
-            </span>
+              <span
+                className="StyledText-sc-1sadyjn-0 bEMqcB"
+                size="small"
+              >
+                ADD LINE
+              </span>
+            </button>
           </div>
           <div
-            className="StyledBox-sc-13pk1d4-0 tkzRw sc-kAzzGY cHmBTk"
+            className="StyledBox-sc-13pk1d4-0 tkzRw sc-cSHVUG iUjjWu"
           >
             <div
               className="StyledBox-sc-13pk1d4-0 joGoBe"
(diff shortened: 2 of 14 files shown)


wgranger:
WIP while I take care of a rebase issue here

shaunanoordin:
Screen Shot 2020-03-16 at 22 02 03

Interestingly, I was able to trigger an error by attempting to add a line to an... unprepared Subject?

1. Go to Project 1765 (staging), Workflow 3085, Group GROUP21, subject 121
2. Note that the panel says "This page does not contain transcription data."
3. click on Add Line button
4. watch as the whole app crashes

To be fair, it was one of those _incredibly wrong_ Subjects (it was a star field, not a page of text!) but it indicates some existence checks need to be set in.

shaunanoordin:
Side note 1: I just realised that the _"Loading..."_ or _"Error: could not find Subject with id = XYZ"_ messages don't display on the Subject viewer UNTIL you move your mouse on the Subject Viewer.

It seems that the span ain't rendering until mouse input comes in, for some bizarre reason.

Side note 2: if I go to Subject A (which has an image), then go back and look for Subject B (which doesn't have an image, and triggers the "could not find Subject with id = B" error message), the Subject Viewer will show the image for Subject A before showing the error message

This may be an indicator that when visiting a new Subject, the "clear/reset previous Subject" logic doesn't kick in _until_ the new Subject data is retrieved. (Expected behaviour: clear/reset previous Subject as soon as a new request for a new Subject is triggered.) Might be worth looking into, as there may be scenarios when a user switches Subjects and doesn't realise they're still seeing the old one.

shaunanoordin:
Oh man, sorry I'm taking so long but I'm going a bit nuts with these UI tests. Some notes _pertinent to the PR_ so far:

1. adding a new line on the Edit page by clicking "Add Line" opens a popup - you can then type in some text, then click "Replace with Selected" to add the line. Henceforth, I shall call these Manual Lines, which is different from the Aggregated Lines.
  - Manual lines have no X-Y coordinates. This is fine.
  - ⚠️ The newly added Manual Line appears at the END of the page, which may not be intuitive.
  - ⚠️ "Replace with Selected" is not an intuitive name for the action button.
  - ⚠️ The "Close" button is way too easy to click. Users will close the popup and not realise they didn't save their changes.
2. [minor] however, after you add a Manual Line, and the popup shows details of the newly created line, you can click "Add Line Below"... which adds another Manual Line.
  - This second "chained" Manual Line is pre-filled with the first Manual Line's text.
  - Mildly confusing, but you know what? No big deal.

shaunanoordin:
I don't know how I missed this until now, but...

☠️⚠️ Side note 3: if you rearrange two Lines, or if you delete one Line and cause the other lines below it to reposition "one step up", **the Lines' raw data gets all screwed up**

For example, let's say we have...

```
Item 1, Aggregated Line: "Hello world" -> raw data ["Herlo world", "hello wrld", "hello world"] 
Item 2, Aggregated Line: "How are you?" -> raw data ["How R you?", "How are u?", "Hw are you?"]
```

...and then let's say we switch the position of Item 1 and Item 2 by click-dragging on the Edit page to re-order the lines. This is what we get:

```
Item 1, Aggregated Line: "How are you?" -> raw data ["Herlo world", "hello wrld", "hello world"] 
Item 2, Aggregated Line: "Hello world" -> raw data ["How R you?", "How are u?", "Hw are you?"]
```

This is obviously very very wrong and confusing AF. If I _refresh the page,_ it'll be fine, so it seems the issue is just that the "raw text" data stores aren't swapped along with the "true final selected text" of the lines.

The only reason I noticed this now is because I was deleting a lot of my newly-created Manual Lines and suddenly noticed that a bunch of them had legit raw text from Aggregation Lines in 'em.

wgranger:
Thanks for the review. I've addressed the issue of showing the right buttons when trying to add a line, as well as an issue where a "viewer" was erroneously able to add a line. 

As for the other issues raised, I've opened a couple issues to handle the:
- [Issue](https://github.com/zooniverse/text-editor/issues/118) with accessing an incomplete transcription
- [Issue](https://github.com/zooniverse/text-editor/issues/120) when rearranging transcriptions

I'd like to address those separately as they weren't raised from any code changes here.

wgranger:
Note: this PR will not currently work until a fix is made in the front end to pass the `Last Modified` header back to the backend. Marking this WIP in the meantime.

wgranger:
Rebased with the patch fix. This should be good to go.

shaunanoordin:
## PR Review (Update)

Yikes! Not sure what's happening here, but there are at least two issues. 1. Attempting to add a line causes a massive crash. And, 2. the crash reporter is also crashing.

Screen Shot 2020-03-27 at 18 35 09

Actions:
1. go to an Edit Subject page, e.g. http://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/12345/subjects/85068/edit
2. Click on the "Add Line" button on the right-hand panel.
3. Type in a value, e.g. "Example Sentence", into the popup then click "Add Line" to confirm.

Expected:
- a new line should be added to the database, and shown on the right-hand panel's list of text lines.

Actual:
- App crashses
- PATCH request to `https://tove-staging.zooniverse.org/transcriptions/85068` returns error 400:
  ```
  status: "400"
  source: null
  title: "Bad Request"
  detail: "You must supply the 'If-Unmodified-Since' header and it must contain the resource's GET 
  request 'Last-Modified' header value."
  ``` 
- The error is thrown across the usual TranscriptionStore's try-catches, but when it gets to `self.error = getError(err)`, there's a problem:
```
  // src/helpers/getError.js
  export default function getError(err = {}) {
    ...
    if (code) {
      ...
      message = err.body.error  // <-- assumes err and err.body exists
    } else {
    ...
  ```

### Status

⚠️ Major error encountered

wgranger:
@shaunanoordin That error should be fixed with the recent rebase I did. I needed to send some headers due to a recent change in TOVE, and the changes are reflected [here](https://github.com/zooniverse/text-editor/pull/132)

Issue #110, opened 2020-03-03
Design Notes March 3


beckyrother:
## Login page
- [x] Bigger
- [x] White box on the right should be height 100%
- [x] Responsive layouts: wide https://projects.invisionapp.com/d/main#/console/17925240/409452614/preview narrow https://projects.invisionapp.com/d/main#/console/17925240/409455403/preview

## Indexes 
**General**
- [x] Clicking back or on the breadcrumb should remember the page in the index the user was on (that is, in this example the list of subjects should be on page 4)
- [x] User dropdown needs more padding around items – should be 44px tall

**Search/Filter**
- [x] Dropdown arrow should be black
- [x] Checkboxes look like they've been stretched? 
- [x] Checkboxes also shouldn't be teal

## Transcribe viewer 
**Main Interface**
- [x] Use Karla Regular (300?) instead of bold (600) for the subject name
- [x] Is there an outline version of the metadata icon? It's quite heavy
- [x] Mark as Approved toggle stroke is too thick, should just be 1px
- [x] Add a checkbox inside the Mark as Approved toggle when the subject has been approved (accessibility)
- [x] Undo icon needs to be a little larger to match the other icons
- [x] All the up/down arrows (More dropdown, etc) need to be reversed so their default state is pointing down
- [x] Original Subject, Transcribed Text, and All Pages labels should be 18px

**Original Subject**
- [x] Zoom controls should have a white background with 50% opacity
- [x] (Change from InVision) Swap zoom controls so Zoom Out is on the far left 
- [x] Clicking on a line in the original subject should highlight the corresponding transcribed line ("As an admin I want to click on a line and view that line’s consensus transcription as well as a grid showing individual (raw) transcriptions.") (#112)
- [x] Switch open and closed circles on lines: in the FER transcriptions we're doing open circles for the start of a line, closed for the end (#105)

**Transcribed Lines** 
- [x] Add some space between `Flag` and `Consensus Score`
- [x] Smaller drag handles on each line

**Transcribed Line Detail**
- [x] Remove the bullet from transcribed details (eg Anonymous)
- [x] Text input should be taller
~- [ ] Clicking anywhere outside the line should close the modal~ 
- [x] Other buttons (toolbar for example) should be disabled while modal is open

**All Pages section**
- [x] Capitalize `p` in `Pages`
- [x] Thumbnails need a hover state
- [x] Let's use the teal as the selected outline since teal is used in the pagination circles

wgranger:
[...]

PR #129, opened 2020-03-23
Interface Design Notes

wgranger:
This PR takes care of remaining items in the "Main Interface" and "Original Subject" section of #110. **Note:** I'm not too worried about the test coverage decreasing here as it's minimal and could also be contributed to removing lines from some refactoring in this PR.

- Adjusted to image tools background and order (Fig. 2)
- Added an "X" and check mark on the mark as approved button
- Switched the up and down arrows for drop buttons
- Changed the font size for header buttons

Fig 1:
Screen Shot 2020-03-23 at 8 52 04 AM

Fig 2:
Screen Shot 2020-03-23 at 8 53 21 AM

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -17,10 +17,11 @@ exports[`Storyshots AggregatedTranscriptions Default 1`] = `
           className="StyledBox-sc-13pk1d4-0 ikTqht"
         >
           <div
-            className="StyledBox-sc-13pk1d4-0 byArxi"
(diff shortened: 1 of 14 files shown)


beckyrother:
Good catch @srallen. 
- Hover states should be the same as focus, yes
- Fallback to come!

beckyrother:
Here's an image to use in case someone doesn't have an avatar: 

[user.svg.zip](https://github.com/zooniverse/text-editor/files/4394964/user.svg.zip)

wgranger:
I've opened a few issues to cover the items brought up in this review. Merging this.

PR #131, opened 2020-03-23
Reparse Extracts after Rearranging

### LINKED ISSUE #120: Rearranging transcription lines bundles reduction
From a [recent PR review](https://github.com/zooniverse/text-editor/pull/107#issuecomment-599791745). Rearranging transcription lines can incorrectly reorder the text. I think the reduction needs to be re-parsed after it is arranged.

wgranger:
Closes #120 

This PR takes care of an issue that emerged from a PR review. Transcription rows were being arranged without reparsing the order of the extracts, which accounts for displaying a user's transcription when the modal is open.


wgranger:
Merging, as this is a one liner and need this for MVP

PR #132, opened 2020-03-24
Send If-modified-since Headers

wgranger:
With recent model locking updates to TOVE (https://github.com/zooniverse/tove/pull/151, https://github.com/zooniverse/tove/pull/171), ALICE now needs to send the last updated time along with each patch. 

This PR is mainly a fix to get patching working again. Another PR with all the model locking bells and whistles will follow.


wgranger:
Merging this to get the fix in and allow other branches to PATCH

PR #135, opened 2020-03-25
Transcription Locking and Unlocking

### LINKED ISSUE #26: Subject Locking Functionality
Handle the dreaded subject locking. We’re going to deal with this through a more straightforward approach, but the frontend needs to have a way to display on the subject data table that X subject is locked. If a user clicks on that subject, they will also see a modal (already-build) showing that the subject is locked.

wgranger:
Closes #26 

Refer to [this doc](https://docs.google.com/document/d/1aBMwZM6X0wf-K0fkE0JPX9h_DHjj_hOBDCJzhJLx3ig/edit) for a bit more backend and frontend info. 

**Use case:**
For the text editor app, we don't want two people to be editing a transcription at the same time. Because of that, transcriptions have a `locked_by` attribute that is set to the person who is currently accessing a transcription. If you try to update a transcription that is `locked_by` another user, the `PATCH` (that occurs when attempting to update) will error, so that is a situation we want to avoid. **Important:** A user can lock a subject for three hours.

Most importantly, after a user finishes editing a transcription, that user should unlock a transcription by hitting an endpoint `/transcriptions/:id/unlock`, and **this is what we attempt to do via this PR**.

**How to Solve:**
Let's think of all the different ways a user might navigate away from a transcription.
- **Closing the Window or Browser:** This should be covered via the `beforeunload` event listener in this PR.  
- **Navigating Back:** This should be handled when the `Editor` component unmounts via the return statement of the Editor's `useEffect` hook in this PR.  
- **Inactivity:** This is the trickiest part. Let's say a user has their window open for over three hours or closes their laptop overnight and expects to open up and keep transcribing. However, upon resuming the work, that transcription has been locked by another user. We want to elegantly message this to the user.

  Because of this, I use the [Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) to check whenever a page changes visibility (triggered by tabbing into or opening/closing a laptop lid). Whenever this happen, we check if three hours (unlocking duration) have passed since the transcription was last accessed. If it has been more than three hours, we GET the transcription again to check if it is still free or has been `locked_by` another user. If the transcription is available, we refresh the access time and begin working for another three hours since it has presumably been re-locked.

### DIFF
--- a/src/components/Modals/SubjectLockedModal/SubjectLockedModal.js
+++ b/src/components/Modals/SubjectLockedModal/SubjectLockedModal.js
@@ -7,7 +7,7 @@ const StyledText = styled(Text)`
   line-height: 1.75em;
 `
 
-export default function SubjectLockedModal({ onBack }) {
+export default function SubjectLockedModal({ lockedBy, onBack }) {
   return (
     <Box
       background='white'
@@ -19,7 +19,7 @@ export default function SubjectLockedModal({ onBack }) {
     >
       <Text size='large'>Subject locked</Text>
       <StyledText>
-        This subject cannot be accessed because <b>Erin Green</b> is currently accessing it.
+        This subject cannot be accessed because <b>{lockedBy}</b> is currently accessing it.
       </StyledText>
       <StyledText>For access, ask them to close their version.</StyledText>
       <Button
--- a/src/components/Modals/SubjectLockedModal/SubjectLockedModal.js
+++ b/src/components/Modals/SubjectLockedModal/SubjectLockedModal.js
@@ -19,7 +19,7 @@ export default function SubjectLockedModal({ lockedBy, onBack }) {
     >
       <Text size='large'>Subject locked</Text>
       <StyledText>
-        This subject cannot be accessed because <b>{lockedBy}</b> is currently accessing it.
+        This subject cannot be accessed because <strong>{lockedBy}</strong> is currently accessing it.
       </StyledText>
       <StyledText>For access, ask them to close their version.</StyledText>
       <Button
--- a/src/screens/Editor/Editor.js
+++ b/src/screens/Editor/Editor.js
@@ -33,6 +33,11 @@ function Editor ({ match }) {
     setResources()
     store.transcriptions.setActiveTranscription()
     undoManager.clear()
+
+    // window.addEventListener('beforeunload', function() {
+    //   store.transcriptions.unlockTranscription()
+    // });
+    return () => store.transcriptions.unlockTranscription()
   }, [match, store])
 
   const disabled = store.aggregations.showModal || store.transcriptions.approved || store.transcriptions.isActive
--- a/src/screens/Editor/Editor.js
+++ b/src/screens/Editor/Editor.js
@@ -4,6 +4,7 @@ import { observer } from 'mobx-react'
 import { withRouter } from 'react-router-dom'
 import AppContext from 'store'
 import { undoManager } from 'store/AppStore'
+import MODALS from 'helpers/modals'
 import Resizer from './components/Resizer'
 import AggregationModal from '../../components/AggregationSettings/AggregationModal'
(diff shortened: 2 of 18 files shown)


srallen:
I have some time to review this tomorrow.

src/components/Modals/SubjectLockedModal/SubjectLockedModal.js
```diff
@@ -19,7 +19,7 @@ export default function SubjectLockedModal({ onBack }) {
     >
       <Text size='large'>Subject locked</Text>
       <StyledText>
-        This subject cannot be accessed because <b>Erin Green</b> is currently accessing it.
+        This subject cannot be accessed because <b>{lockedBy}</b> is currently accessing it.
```
srallen, line 22: I think [`<strong>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) is a better element to use rather than `<b>`. The `<b>` element is one of those older HTML 4 elements that is discouraged now to use because it's not semantic. The other option would be to wrap this in a span and use CSS to style it.

src/screens/Editor/Editor.js
```diff
@@ -21,19 +21,40 @@ function findLocations(subject) {
...
+      const shouldRecheck = new Date() > recheckTime
+      if (shouldRecheck) {
+        accessTime = new Date()
+        recheckTime = new Date(accessTime).setHours(accessTime.getHours() + 3)
+        store.transcriptions.checkIfLocked()
+      }
+    }
+
+    window.addEventListener('beforeunload', store.transcriptions.unlockTranscription)
+    window.addEventListener('visibilitychange', handleTimeCheck)
```
srallen, line 50: Just for my own clarity, these two actions of setting the resources and active transcription and then handling the time check and event listeners are in the same effect because you want these both to happen on mount and unmount?

wgranger:
Interesting. A couple things I need to look at here. Upon further inspection, TOVE might be overwriting the `locked_by` with the most recent user getting the transcription, whereas it should instead leave the transcription locked. However, I've also noticed a couple other items to address.

Also, I'll open an issue on a design audit in Firefox, because the browser is displaying things a bit differently from Chrome.

wgranger:
Marking this WIP until we get a backend fix in

src/screens/Editor/Editor.js
wgranger: I only want them to happen on mount. Anything wrapped below in the `return () => { ... }` block will execute on unmount, which is removing the event listeners and unlocking the transcription. However, in the case the window is closed explicitly (clicking the browser 'X' button) without unmounting the component (eg. navigating to another part of the page) then this `beforeUnload` listener would kick in.

wgranger:
@srallen This is ready for another review after https://github.com/zooniverse/tove/pull/182 merging

src/screens/Editor/Editor.js
srallen: Thanks for the order of operations explanation!

src/store/TranscriptionsStore.spec.js
```diff
@@ -276,6 +284,23 @@ describe('TranscriptionsStore', function () {
...
+      it('should unlock a transcription', async function () {
+        await transcriptionsStore.unlockTranscription()
+        expect(patchToveSpy).toHaveBeenCalledWith(
+          '/transcriptions/1/unlock',
+          {"headers": {"If-Unmodified-Since": "Mon, June 31, 2020"}}
+        )
+      })
+
+      it('should return if the transcription is lockedByCurrentUser', function () {
+        expect(transcriptionsStore.lockedByCurrentUser).toBe(false)
```
srallen, line 301: Is there a test to check the conditions when this is true? Maybe this evaluates true in the it statement for `should show when the transcription is locked`?

wgranger: There is not. It will take a bit of setup to make this check but I don't mind adding it.

PR #143, opened 2020-03-30
Button hover state

### LINKED ISSUE #139: Add Hover State to Buttons
The header buttons do not currently have a hover state. This arose from a recent [PR review](https://github.com/zooniverse/text-editor/pull/129#issuecomment-605109869), and the decision is that hover states should be the same as focus states.

wgranger:
Closes #139 

This PR gives hover states to all buttons. Visually, hover states should be the same as focus states and they should not occur when a button is disabled.

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -141,7 +141,7 @@ exports[`Storyshots AggregationSettings Default 1`] = `
               </span>
               <button
                 aria-label="Close Aggregation Settings"
-                className="StyledButton-sc-323bzc-0 jfbKsv"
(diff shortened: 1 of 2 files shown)


beckyrother:
Sorry is the hover state an underline at the moment or just the focus box?

srallen:
Right now it's just the focus outline. I'm suggesting in my review that Will use the `PlainButton` we've already built with the underline styles, but I need to publish it to enable that to happen.

beckyrother:
Good call, should probably add an underline. Thanks!

wgranger:
Cool. Looks like this is good to go for now. I'll open an issue to import `PlainButton` when that is available. Maybe it can be added when we do another audit of the site after alpha/beta testing.

PR #144, opened 2020-03-30
Use CSS Helpers to Avoid Memory Leaks

### LINKED ISSUE #141: Use css Helper for Styled Components
Memory leaks could occur when using styled components without the CSS helper. This arose from a recent [PR review](https://github.com/zooniverse/text-editor/pull/129#pullrequestreview-383029947). There are a few components in this repo that will need this attention.

wgranger:
Closes #141 

This PR uses the styled-components CSS helpers to avoid an issue with memory leaks.

### DIFF
--- a/src/components/Badge/Badge.js
+++ b/src/components/Badge/Badge.js
@@ -32,7 +32,7 @@ const DropLink = styled(Link)`
 `
 
 const StyledBox = styled(Box)`
-  background: ${props => props.disabled ? '#D8D8D8' : 'inherit'}
+  ${css`background: ${props => props.disabled ? '#D8D8D8' : 'inherit'};`}
   :hover {
     background: #D8D8D8;
   }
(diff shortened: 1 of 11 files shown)


src/components/Badge/Badge.js
```diff
@@ -26,7 +26,8 @@ const DropItem = styled(Button)`
 `
 
 const DropLink = styled(Link)`
```
shaunanoordin, line 28: `DropLink` looks fine, but what about `StyledBox`, just below this code block?

```
const StyledBox = styled(Box)`
  background: ${props => props.disabled ? '#D8D8D8' : 'inherit'}
  :hover {
    background: #D8D8D8;
  }
```

wgranger: Oops! I missed that one

wgranger: Looks like that came in when this branch was rebased. Good catch.

PR #154, opened 2020-04-17
Fix bug when searching by Zooniverse id

wgranger:
The search modal was incorrectly pinging the TOVE db for `subject_id`, which doesn't exist. Instead, `id` should work here since TOVE transcription IDs map exactly to Panoptes subjects


wgranger:
Merging. This is such a small fix and I'd like to get it in for MVP

PR #158, opened 2020-04-30
Rearrange Pages By Slope Values

### LINKED ISSUE #95: Sort transcriptions by slope value
According to the [filmstrip viewer](https://projects.invisionapp.com/d/#/console/17925240/401676674/preview), transcriptions need to be further subdivided within each frame by their rotation/slope.

wgranger:
Closes #95 

[Invision Design](https://projects.invisionapp.com/d/#/console/17925240/371813757/preview) - note the blue outline is outdated

In a nutshell, this PR does two things:
- Shows all slope values for a given page, allows users to select different values, and denotes them when not 0 degrees
- Allows users to rearrange the order of these slope frames and saves the resulting order

There were many gotchas when writing this PR, but they _should_ all be resolved:
- Will the page order be rearranged to the original values when choosing to "edit aggregation settings" back to the original values? (default values under DB SCAN)
- Will the transcription retain frame order when reloading the transcription
- Will the UI show when frames with the same page number are grouped together?
- Can a user correctly "undo" a change in frame order? For this to work correctly, moving frames should only commit to the state tree when the frame drag ends (not with each move over a frame). 

Things to keep in mind during review:
- The dependency array in React's `useEffect` hook, which determines when the hook triggers, does **not** notice changes when the order of items in an array change. Ex: If my array is `[frame0, frame1, frame2]` and it changes to `[frame1, frame2, frame2]`, the hook will not register a change and `useEffect` will not fire. Thus, I had to write a custom `usePrevious` hook and check the array values were different with an `isEqual` helper function.

- With mobx state tree, using `detach` to remove unused nodes is imperative. Let's say I separated `frame0` into `frame0` and `frame0.1` by moving half of that frame to the very end of all the frames. Next, I selected that `frame0.1` at the end of the page. Finally, I decide to "edit aggregation settings" and revert back to the original reduction. When I do this, there is a split second where `frame0.1` still attempts to render, even though the reaggregation combines it back into the original `frame0` value. Mobx state tree would choke on this because `frame0.1` doesn't actually exist in the tree anymore, but it is still erroneously being referenced.

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -520,7 +520,7 @@ exports[`Storyshots FilmstripViewer Default 1`] = `
         </button>
       </div>
       <div
-        className="StyledBox-sc-13pk1d4-0 ekjAwv"
+        className="StyledBox-sc-13pk1d4-0 ilHtzj"
       />
     </div>
   </div>
--- a/src/components/AggregatedTranscriptions/components/TranscriptionTable/TranscriptionTable.js
+++ b/src/components/AggregatedTranscriptions/components/TranscriptionTable/TranscriptionTable.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Box, Text } from 'grommet'
 import styled from 'styled-components'
-import { arrayOf, bool, shape, func } from 'prop-types'
+import { arrayOf, bool, number, shape, func } from 'prop-types'
 import { observer } from 'mobx-react'
 import TranscriptionTableRow from './TranscriptionTableRow'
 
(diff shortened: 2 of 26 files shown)


shaunanoordin:
The automated checks have borked at commit dacb348 ; let me see what's up

shaunanoordin:
Oh, it's just coveralls. Move along, nothing to see here.

wgranger:
Yeah, coverage decreased slightly, but I did also remove a good bit of lines with this code. I still feel like the relevant items are covered and don't want to spend too much of my effort getting to 100% coverage.

shaunanoordin:
Screen Shot 2020-05-13 at 23 00 04

I AM Darke Shard, goddammit! I just opened the Subject and I got this. 🤔

wgranger:
Page angles are rounded to the nearest 10 degrees. This is something that can be played around with [this](https://github.com/zooniverse/text-editor/pull/158/files#diff-f851089418ff362b03294616db1ff0c1R11), but that can likely be tweaked during the initial review.

There is a `frame_order` key on the transcription object from TOVE that denotes when the frame order has changed. The format may be `['frame0.0', 'frame0.1']` where the `.0` and `.1` corresponds to the `slope_label` of each transcription (not the actual value). Info from the reduction docs is [here](https://docs.google.com/document/d/1gG-1jnSvRN6gUWwLNv6TJxM-ujN5hupTGaWuxNQ70R4/edit#).

PR #161, opened 2020-05-14
Test for Transcription Locked By Different User

wgranger:
I believe this should take care of the subject locking issue that was first identified in #158 

I believe `null` values were being interpreted as a transcription being locked. The check was:

- See if the current transcription's `locked_by` equals the current user
- If it does not equal, show the locked modal

What if `locked_by` is `null`, meaning it is available/unlocked?
- Check if `null` equals the current user
- It isn't equal, show the locked modal

This doesn't make sense!

This PR should rectify that issue by checking instead if the transcription is locked by a **different** user and responding accordingly. I'm not positive it should work, as I wasn't able to replicate the issue, but it should take care of what I believe the issue to be.

### DIFF
--- a/src/store/TranscriptionsStore.js
+++ b/src/store/TranscriptionsStore.js
@@ -528,8 +528,8 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
   },
 
   get lockedByDifferentUser () {
-    const user = getRoot(self).auth.userName
-    return self.current.locked_by && self.current.locked_by !== user
+    const login = getRoot(self).auth.user.login;
+    return self.current.locked_by && self.current.locked_by !== login
   },
 
   get isActive () {
--- a/src/store/TranscriptionsStore.js
+++ b/src/store/TranscriptionsStore.js
@@ -102,7 +102,7 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     const response = yield client.get(`/transcriptions/${self.title}`)
     const resource = JSON.parse(response.body)
     const lockedBy = resource.data.attributes.locked_by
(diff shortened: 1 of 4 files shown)


src/store/TranscriptionsStore.js
```diff
@@ -439,9 +439,9 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     return count;
   },
 
-  get lockedByCurrentUser () {
+  get lockedByDifferentUser () {
     const user = getRoot(self).auth.userName
```
shaunanoordin, line 443: This is very odd.

For my staging user account, my details are...

```
{
  login: 'darkeshard',
  display_name: 'Darke Shard'
}
```

The `ToveTranscriptionResource.lockedby` value keeps track of the user's **login,** but `auth.userName` is pulling the user's **display name.**

Question: does the account you use for testing have login === display_name?

wgranger: Welp. You might've found the issue. Let me spin this up and take a look.

wgranger: I think that's the fix we need. Why do we have `login` and `display_name`?

wgranger:
@shaunanoordin Thanks for finding the issue here! Turns out the fix was easier than I imagined. In the future, will have to remember we have both the `login` and `display_name` properties, the latter of which can be changed by the user.

wgranger:
I believe the only item that needs changing is in the `checkIfLocked` function, so I went ahead and made the switch. I don't think you would be able to trigger that naturally unless someone came in and accessed the transcription while your browser was asleep for some time and your lock timed out.

PR #164, opened 2020-05-20
Adjust Metadata Modal

### LINKED ISSUE #163: Clean up the metadata container
This should be an easy fix. Looks like something got lost along the way and the metadata headers on the modal aren't populating with any subject info. Also, any header key should be removed if the value is an empty string in the case of the image100, image101, etc. items below.

Screen Shot 2020-05-19 at 9 48 09 AM

wgranger:
Closes #163 

Upon an initial audit, I noticed the metadata needed cleaning. More info in the associated issue, but this also depended on bringing a bit more transcription data into the modal.

### DIFF
--- a/src/components/EditorHeader/components/MetadataButton/MetadataButton.js
+++ b/src/components/EditorHeader/components/MetadataButton/MetadataButton.js
@@ -47,8 +47,6 @@ function MetadataButton({
   const targetEl = React.useRef(null)
   const [isOpen, toggleDrop] = React.useState(false)
 
-  console.log('TRANSCRIPTION', transcription);
-
   const { pages, status, transcribed_lines } = transcription
 
   return (
--- a/src/components/EditorHeader/components/MetadataButton/MetadataButton.js
+++ b/src/components/EditorHeader/components/MetadataButton/MetadataButton.js
@@ -98,13 +98,15 @@ function MetadataButton({
                 </colgroup>
                 <TableBody>
                 {metadata && Object.keys(metadata).map((key, i) => {
+                  const value = metadata[key]
+                  if (!value) return null
                   return (
                     <TableRow key={`METADATA_VALUE_${i}`}>
                       <TableCell>
                         <CapitalText>{key}</CapitalText>
                       </TableCell>
                       <TableCell>
-                        <StyledText>{metadata[key]}</StyledText>
+                        <StyledText>{value}</StyledText>
                       </TableCell>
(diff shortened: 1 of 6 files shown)


src/components/EditorHeader/components/MetadataButton/MetadataButton.js
```diff
@@ -38,18 +38,17 @@ const StyledFontAwesomeIcon = styled(FontAwesomeIcon)`
...
-  score,
-  status,
-  transcribers
+  transcriberCount,
+  transcription
 }) {
   const targetEl = React.useRef(null)
   const [isOpen, toggleDrop] = React.useState(false)
 
+  const { pages, status, transcribed_lines } = transcription
```
shaunanoordin, line 50: 1. can `transcription` be `null`? If yes, you might want to make this `... = transcription || {}` or something. (Not sure if null will auto-overwrite the values in defaultProps, mind you.)

2. I assume if you ever have `transcription`, that object will (since it's coming from the API) always have pages, status, transcribed_lines... and even if it didn't, the worst that can happen is you get status = undefined. OK. 👌

src/helpers/gsCountFromExtracts.js
```diff
@@ -0,0 +1,9 @@
+export default function gsCountFromExtracts(extracts) {
+  let count = 0;
+  extracts.forEach(extract => {
+    const values = Object.values(extract.data)
```
shaunanoordin, line 4: Just to confirm: `extract.data` will always exist?

src/components/EditorHeader/components/MetadataButton/MetadataButton.js
wgranger: Good call. Def won't hurt to add a fallback value here

src/helpers/gsCountFromExtracts.js
wgranger: Should prob add an existence check

shaunanoordin:
I hereby grant you this seal of approval.

Seal GIF

wgranger:
Updated tests and noticed I should remove the check here https://github.com/zooniverse/text-editor/pull/164/commits/a8d3149ef29209d40cba80af163046bf364760e1#diff-f84f02ecd29a62a11c61ba200c1028f0R50

There's no way for that line to fire. If the transcription prob was undefined the default props declaration would kick in and provide values.

PR #166, opened 2020-06-02
Fix Subjects Page Pagination Issue

wgranger:
This fixes an issue with pagination on the transcriptions container. What was once appearing as this:

Screen Shot 2020-05-29 at 11 27 24 AM

Should now look like this (apologize for the quick scrolling in the video):

Paginate

### DIFF
--- a/src/components/StepNavigation/StepNavigation.js
+++ b/src/components/StepNavigation/StepNavigation.js
@@ -21,8 +21,15 @@ const StyledButton = styled(Button)`
 const StyledRadioButtonGroup = styled(RadioButtonGroup)`
   position: relative;
   > label {
+    display: flex;
+    flex-direction: column;
+
     > div {
-      margin-right: 10px;
+      margin: 0 0.25em;
+    }
+
+    > span {
+      font-size: 0.75em;
     }
   }
 `
@@ -35,17 +42,19 @@ class StepNavigation extends React.Component {
   }
 
   render () {
-    const { activeStep, disabled, setStep, steps } = this.props
+    const { activeStep, disabled, setStep, showLabel, steps } = this.props
     if (steps && steps.length > 1) {
       const nextStep = activeStep + 1
       const prevStep = activeStep - 1
       const options = steps.map((step, index) => {
+        const showPage = showLabel && !isNaN(step)
         // We can't just use index for the value
         // because Grommet is using indexes internally as keys and this will error with a duplicate key
-        const value = `step-${index}`
+        const value = `step-${showPage ? step : index}`
         return {
           disabled,
           id: value,
+          label: showPage ? (step + 1).toString() : null,
           value
         }
       })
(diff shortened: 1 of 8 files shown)


src/components/StepNavigation/StepNavigation.js
```diff
@@ -28,44 +34,65 @@ class StepNavigation extends React.Component {
...
-            gap='none'
-            name='step-selectors'
-            onChange={this.onChange.bind(this)}
-            options={options}
-            value={`step-${activeStep}`}
-          />
-          <StyledButton
+
+          <Box direction='row'>
+            {activeStep > 2 && <Text>...</Text>}
```
shaunanoordin, line 70: [minor] For readability, this can be rewritten with `const hasMoreThan2Steps = activeStep > 2`

src/screens/SubjectsIndex/SubjectsPageContainer.js
```diff
@@ -7,8 +7,21 @@ import { EDIT_PATH } from 'paths'
...
 import { columns } from './table'
 
+function slicePages(page, totalPages) {
+  const allPages = Array.from(Array(totalPages).keys())
+  let leftPage = page - 2 < 0 ? 0 : page - 2
+  let rightPage = leftPage + 5
+  if (totalPages > 5 && rightPage > totalPages - 1) {
+    leftPage = totalPages - 5
+    rightPage = totalPages
+  }
```
shaunanoordin, line 17: [minor] You might want to make some of these magic numbers into constants, e.g. `const MAX_NUMBER_OF_PAGES_SHOWN_IN_PAGING = 5`

wgranger:
@beckyrother This shows the current page as well as two pages to the left and right of the current page, except in the case you are near the first or last page.

If you are near the first page it would show:
1 **2** 3 4 5...

Near the last page, assuming last page is 30:
...26 27 28 **29** 30

beckyrother:
@wgranger Yes, I think it would be helpful to know how many pages total there are, if that doesn't add a ton of extra work. Agreed that that would make double arrows unnecessary. 

So if you were on the second page it would show:
1 **2** 3 4 ... 30

wgranger:
@beckyrother Was an easy enough fix. How about this?

Screen Shot 2020-06-05 at 3 44 18 PM

beckyrother:
Looks great! Approved here 👍

wgranger:
@shaunanoordin This is ready for a re-review after a design update.

src/theme.js
```diff
@@ -104,6 +104,17 @@ const theme = {
       background: 'rgba(225, 225, 225, 0.75)'
     }
   },
+  radioButton: {
+    extend: props => css`
```
shaunanoordin, line 108: [super minor] does this need to be a `props => {}` function? The CSS values seem pretty static. But hey, it works, and the `css()` call is there to prevent ye olde memory leakes, so all good. 👌

PR #168, opened 2020-06-03
Fix Issues with Extracts

wgranger:
There were a few bugs with extracts that this PR aims to fix:

- Add extracts lines back to the subject (erroneously removed in #158)
- Fix several issues with adjusting frames with sub-keys (ex: `frame0.0`, `frame0.1`, etc.)
    - Allow lines to be added
    - Allow lines to be deleted
    - Allow lines to be reordered


wgranger:
Merging this now so we can begin testing

PR #171, opened 2020-06-09
Adjust Subjects Page Container Headers

wgranger:
This is a quick change to fix the headers of the subjects page container based on a recent call. To match the following:
Screen Shot 2020-06-09 at 3 56 19 PM

Original headers:
Screen Shot 2020-06-09 at 4 00 16 PM

The new header reads "Low Consensus Lines," whereas before it read "Consensus Score," but was pretty vague and was basically a fraction of `low_consensus_lines/transcribed_lines`. The following column already includes "Transcribed Lines," so listing it twice was redundant.


wgranger:
Merging as this is such a small change that we need to test out in a deploy.

PR #172, opened 2020-06-11
Reaggregation Spinner

wgranger:
When editing aggregation settings, there really needs to be a loading spinner to block interactivity so a user doesn't try to edit a reduction while it is being reaggregated.

Screenshot:
Screen Shot 2020-06-11 at 2 12 46 PM

And gif:
loading

I removed the animation on the modal appearance, because that was a bit jarring. Now the modal appears instantly. 

Only remaining question is: **should we provide a button on the modal?** Cancelling a request isn't the easiest thing to implement, and I'd want to discourage doing anything besides waiting for the call to finish.

However, is it bad UX to disable all interaction on the screen without a way to exit?


wgranger:
@beckyrother Here's another pass:
Screen Shot 2020-06-11 at 3 47 44 PM

wgranger:
Good call. I'll add some error logic to this.

wgranger:
I think this should handle the error handling. Was also making sure an error appearing would not interfere with the "undo" history.

PR #174, opened 2020-06-25
Show Zoom/Rotate Tools with Active Line

wgranger:
This PR shows the zoom and rotate tools at all times. Previously, the tools were disabled when the modal was open to edit a line.


wgranger:
Merging this as it is a small fix that changes a single line

Issue #165, opened 2020-05-27
ALICE updates (to be made after testing)


snblickhan:
- [ ] on Help page, in section beginning with "To ensure correct indexing within ALICE": change **Group_ID** and **Internal_ID** to **group_id** and **internal_id** (remove caps)

nciemniak:
Transcription remains locked when I am viewing the transcription and close the browser window. Update such that transcription unlocks upon window closing.

wgranger:
@mrniaboc 
1) These smaller subject IDs are expected if you were looking at the "TEST" group on ASM. This is because we inserted duplicate transcriptions/subject IDs into TOVE, but intentionally gave these duplicate subjects lower IDs (that we acknowledge don't match their true Zooniverse IDs) since they needed to be unique. This is a temporary fix.

trouille:
In Chrome v83 on macbook pro mojave 10.14.3
Minor: flag and mark as read icons are quite close and my cursor would do funny things (flickering) when going to click on one of them (maybe because of proximity to other)?

wgranger:
@trouille There are a few known issues that still need to be sorted out with the undo button.

wgranger:
> * [ ]  When viewing zoomed-in subjects and clicking on a line further down the page (i.e. outside the viewing field), the zoomed-in image does not automatically 'jump' to the selected line. Not sure this requires action/updating on our part (unless there's a simple fix?), but I can imagine users reporting it as a bug so thought I'd flag.

@snblickhan I believe the BPL team noted this as well. This was not part of the original functionality of ALICE. If I'm understanding this correctly, it is not the quickest fix and will probably take at least a couple days to do this correctly.

beckyrother:
Deleting frames: We discussed on #177 that collapsing the filmstrip viewer might be unnecessary. If that's the case, we should use that space for the Delete Page button, which displays Xs next to the pages and allows the user to delete a page. A confirmation box appears before action is taken. https://invis.io/HAVICPG872T#/371813754_Filmstrip

wgranger:
> * [ ]  When moving from subject view to group view subject remains listed as LOCKED until page is refreshed. (Possible lag in unlocking -- call being made to unlock, but not instantaneous in TOVE?)
> 

It looks like the "unlock" call is made to TOVE on both page close and going back to the transcriptions index. There is a rare state where:

- Editing subject 1234 (locked by you)
- Click breadcrumb to go to transcriptions page (unlock transcription)
[...]

PR #175, opened 2020-06-26
Disable Filmstrip Thumbnail Reordering with Approved Subjects

wgranger:
This PR addresses an item in #165. Once a subject is approved, volunteers shouldn't be able to make any edits. That includes rearranging pages. There is also a bit of refactoring here. The PR covers several items:

- When transcription is approved, disable rearranging pages
- When subject is not approved, show a "move" cursor when hovering over a filmstrip thumbnail
- Refactoring: Abstract out all "status" strings into a single file.


wgranger:
Updating tests to get coverage back up

PR #180, opened 2020-07-02
Fix Logout Crash on Editor Page

### LINKED ISSUE #179: Error When Logging Out
From Slack:

ERROR 1: CRASHING WHILE LOGGING OUT
- Logging out while on Subject Editor page causes a crash.
- Console Error: Cannot read property 'login' of null, at Object.get lockedByDifferentUser 
- Presumably, this has something to do with the checks to see if a subject is locked by another editor.
- LOW PRIORITY

wgranger:
Closes #179 

This PR fixes the issue mentioned above and also handles a few more issues:

- Refactors the code by changing `lockedByDifferentUser` to `lockedByCurrentUser`. 
    - I know this has gone both ways in the past, but I now think `lockedByCurrentUser` makes for a cleaner conditional
- Refactor some of the testing in the `TranscriptionsStore`
    - Initially, that store set up the primary "fetch transcription" test as retrieving a transcription you were locked out of. This is more of an edge case and should be separated into a different `describe` block.

### DIFF
--- a/src/store/TranscriptionsStore.js
+++ b/src/store/TranscriptionsStore.js
@@ -486,8 +486,10 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
   }
 
   const unlockTranscription = flow(function * unlockTranscription() {
-    const { client } = getRoot(self)
-    yield client.patch(`/transcriptions/${self.current.id}/unlock`, { headers: { 'If-Unmodified-Since': self.current.last_modified } })
+    if (self.lockedByCurrentUser) {
+      const { client } = getRoot(self)
+      yield client.patch(`/transcriptions/${self.current.id}/unlock`, { headers: { 'If-Unmodified-Since': self.current.last_modified } })
+    }
   })
 
   function updateApproval(isChecked) {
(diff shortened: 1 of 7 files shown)


src/store/TranscriptionsStore.js
```diff
@@ -545,9 +547,12 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     return count;
   },
 
-  get lockedByDifferentUser () {
-    const login = getRoot(self).auth.user.login;
-    return self.current.locked_by && self.current.locked_by !== login
+  get lockedByCurrentUser () {
+    const login = getRoot(self).auth.user && getRoot(self).auth.user.login;
+    if (login && self.current) {
+      return self.current.locked_by && self.current.locked_by === login
```
shaunanoordin, line 553: Minor comment, mostly to test the logic capabilities of my brain:

Isn't the existential check of `self.current.locked_by &&` unnecessary, technically? I assume the check is there to ensure `self.current.locked_by` isn't null, but you'll never get the unwanted true value of `self.current.locked_by (null) === login (null)` since `login` can never be null/undefined due to the earlier `login &&` check.

That said, don't change this line of code - in the context of our JS coding style, your code is more readable this way.

wgranger: Yep, that's exactly correct. It does seem a bit strange to have `self.current.locked_by && self.current.locked_by === login`, but it's there to guard against null cases.

PR #183, opened 2020-07-06
Fix Pan Tool Acting Clunky

wgranger:
Fixing an issue from #165 

The pan tool doesn't perform as well as other transcription frontends (ASM/Scribes of the CG). I noticed the main difference in the setup is the use of an `InteractionLayer` here to more closely mimic the monorepo classifier. 

However, there is really no need for an interaction layer here, and it merely adds another SVG within an SVG when the parent SVG is the one that should handle the events. The tests are also much nicer here.


wgranger:
The coverage drop here is due to removing all of the `InteractionLayer` components, which all had high test coverage. 

I'm not too concerned with the "drop" in coverage here as `SVGView` is 100% covered, and that is the main component changed here.

wgranger:
Oh yes, `e.preventDefault()` does make drag off much nicer. Making changes now.

PR #184, opened 2020-07-06
Press Escape to Close Selected Transcription

wgranger:
Taken from #165 

This PR closes the `LineViewer` (Selected transcription) modal when pressing the escape key. Tests are included.

### DIFF
--- a/src/components/LineViewer/LineViewer.js
+++ b/src/components/LineViewer/LineViewer.js
@@ -61,6 +61,18 @@ function LineViewer ({
     if (isNewLine) closeModal()
   }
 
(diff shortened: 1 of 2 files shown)


wgranger:
@beckyrother 

Should this be denoted anywhere? This PR lets "escape" close the selected transcription line modal, but there's no signage anywhere telling the volunteer about the quick key.

beckyrother:
It could be included in the help documentation, but I don't think we necessarily need to signpost anywhere as it's a fairly common UX pattern.

wgranger:
Merging. This is a small PR needed for beta.

PR #186, opened 2020-07-07
Reset Image Position on Subject Change

wgranger:
From Slack:

PAN/ZOOM DOESN'T RESET ON CHANGING SUBJECTS
- Pan/Zoom resets correctly when changing pages of a Subject, but not when changing to a different Subject.

Image position (scale, translateX, translateY) now changes on `Editor` unmount, or whenever someone navigates away from the editor page.


wgranger:
Merging - this is a small fix

PR #188, opened 2020-07-10
Rapid Save Bug

### LINKED ISSUE #187: Possible Error 409 from Tove if submitting too many requests at a time
## Potential "Save Data" Bug

1. It's possible to get an Error 409 from Tove if you send too many POST/PUT/PATCH/whatever requests in one go.
2. Alice does not know how to handle an Error 409.

Screen Shot 2020-07-07 at 17 43 34

### Dev Notes

Replication: (At least, what I did to consistently get the error above)

- Go to http://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/12345/subjects/85068/edit
- Observe first transcribed line is "Boston March 3 - 1837", second line is "Friday noon"
- Click & drag to move "Boston March 3 - 1837" to second line (action 1)
- Click on "Boston March 3 - 1837" to open popup (action 2)
- Write new text: "BOSTON MARCH 3 - 1837". Click "Replace with Selected" (action 3)
- Click "Close" (action 4)
- Now RAPIDLY CLICK ON THE UNDO BUTTON 💥 💥 💥 💥 💥 💥 💥 💥 
- Expected: all 4 actions should be undone and saved to Tove
- Actual: it looks like actions 1 to 4 are undone in on the front end. Then, after a few seconds of blissful silence, the whole app crashes.
  - Console error message: `PATCH https://tove-staging.zooniverse.org/transcriptions/85068 409`
  - Further adventures in console error messages:
    ```
    getError.js:25 Uncaught (in promise) TypeError: Cannot read property 'error' of undefined
        at getError (getError.js:25)
        at TranscriptionsStore.js:309
    ```
  - After reloading, it's apparent that actions 2 to 4 are undone, but action 1 (rearranging to second line) remains saved to the database.

This is likely a race condition where two PATCH requests are occurring near-simultaneously (for action 3 and action 1, to be specific) before the first one could be fully processed by the database.

Other ways to replicate the problem:
- Shuffle the order of the first few lines of any Subject REALLY REALLY FAST.

### Status

Major problem (crashes application and leaves state of Subject transcriptions in a potential limbo) but possibly low probability of happening? I mean, what maniac would perform so many changes in one go?

...but if Tove ever takes longer than it currently does to confirm a PATCH request (e.g. user has a slow network, or the payload is megahuge), or if users act faster than the PATCH request can resolve, then we've got a problem.

Note that one sensible solution is to make the the "save changes to database" system action to trigger only once every X seconds (and only if there's no active "save changes to database" action currently happening) but that would require a major rethink of the ALICE design.

wgranger:
This is initially a RFC because tests are failing as I can't find a way yet to yield to a call inside of `setTimeout`. Hoping for some feedback. Otherwise, this closes #187 by adding PATCH requests to a queue whenever they occur and only executing the newest patch after a certain amount of time has passed.


wgranger:
This needed some test reorganization for the jest [timer mocks](https://jestjs.io/docs/en/timer-mocks) to work. To do this, I separated a couple troublesome tests into a separate `TranscriptionsStore_async.spec.js` file, which does clean up the bloated tests a bit.

I'll admit, I'm not quite sure _why_ the tests don't work in the same file, but I believe it has to do with the `jest.runAllTimers()` acting on all tests in the file instead of each test block. Separating the files and using `jest.runOnlyPendingTimers()` does help. 

While I could spend more time trying to figure out the peculiarities of the timer mocks, I'm timeboxing myself and need to move on to other work.

src/store/TranscriptionsStore.js
```diff
@@ -529,6 +545,7 @@ const TranscriptionsStore = types.model('TranscriptionsStore', {
     setParsedExtracts,
     setTextObject,
     setTranscription: (transcription) => undoManager.withoutUndo(() => setTranscription(transcription)),
+    test,
```
shaunanoordin, line 548: Is this meant to be here? It's crashing the app when I try to compile: `ReferenceError: test is not defined`

src/store/TranscriptionsStore.js
```diff
@@ -20,6 +20,9 @@ const Extension = types.refinement(types.map(Frame), snapshot => {
   return Ramda.all(Ramda.startsWith('frame'), Ramda.keys(snapshot))
 })
 
+let patchQueue = []
```
shaunanoordin, line 23: Wait a goshdarn second... is `patchQueue` just floating around as a global variable, instead of an array in the TranscriptionStore's model?

shaunanoordin:
## Recommendations

Here are my thoughts: I think the best way to implement the queue _isn't_ to use a timer to _delay_ patch jobs. Instead, it's to use a timer to _periodically check_ when patch jobs are complete, so the next one can be sent. 

```
const TranscriptionsStore = types.model('TranscriptionsStore', {
  patchQueue: types.array(types.ihavenoidea),
  patchTimer: types.ihavenoidea
  ...
}).actions(self => {
  ...

  function enqueuePatch (query) {
    
    // (STEP 1) Add stuff to the (end of the) queue
    self.patchQueue.push(query)

    // (STEP 2) There should only be one active queue system at a time.
    if (!self.patchTimer) {
      self.patchTimer = setTimeout(
        self.processTheQueue,  // Good luck ensuring the 'self' works properly with this, usually I use myFunction.bind(this)
        0  // Start immediately and asynchronously
      )
    }
  }
                            
  function processTheQueue () {
    
    // (STEP 3) Perform a patch action... ONLY if there's no active patch action going on!
    if (self.patchQueue.length > 0
        && self.asyncState === ASYNC_STATES.READY) {
      
      self.asyncState = ASYNC_STATES.PATCHING  // or LOADING, I guess. It seems that .patchTranscription() doesn't have a status for when it's processing a job.
      
      const oldestPatch = self.patchQueue[0]
      self.patchTranscription(oldestPatch)  // See (!!!) note below

      // TODO: Remove the patch from the queue... but ONLY if it was successful!
      //   self.patchQueue.shift()  // .shift() to remove first item in queue

      // TODO: add in error handlers here.
      // e.g. if patchTranscription throws an error, just SHUT DOWN EVERYTHING.
      //   self.patchQueue = []
      //   alert('ooohhh shiiiiiii-')
      //   throw('something effed uuuppp')
      // Do NOT fail silently! 
    }
    
    // (STEP 4) As long as there are items in the queue, check again in 1 second.
    if (self.patchQueue.length > 0) {
      self.patchTimer = setTimeout(
        self.processTheQueue,
        1000  // The cycle begins anew!
      )
    } else {
      // (STEP 5) If there are no more items in the queue, shut it down.
      
      clearTimeout(self.patchTimer)  // This cleans up any active timeouts which is, uh... unlikely. 
      self.patchTimer = undefined  // This makes sure (STEP 2) above knows it needs to spin up a new queue.
    }
  }
```

The code above has a lot of gotchas, but I hope it properly illustrates my recommendation for a queue system.

(!!!) Note: one of the biggest gotchas for me is that I'm not very familiar with how flow() generators work, so I'm not sure whether `self.patchTranscription(oldestPatch)` would block processing until it's complete, (i.e. synchronous or async?) and how you'd chain success/failures to it.

If we were doing promises, I'd have written something like...

```
self.patchTranscription(oldestPatch)
.then(() => { self.patchQueue.shift() })
.catch(() => { alert('ooohhh shiiiiiii-') })
```

but when it comes to flow() generators, I'm all 🤷‍♂️ ❓

src/store/TranscriptionsStore.js
wgranger: Ah no. Have to clean that up

wgranger:
@shaunanoordin  One note about how this code works that I probably should've made at the beginning.

Every time we save a transcription, the _entire_ transcription is sent over to be updated: the JSON text blob, attributes, etc. Because of that, if we send two quick patches in at once, it's only imperative the _most recent_ patch is actually applied, since changes in the first patch will be included in the second patch as well.

Keeping that in mind, I'll explain what is happening here:
```
      setTimeout(() => {
        const mostRecentPatch = patchQueue.pop()
        if (mostRecentPatch) {
          self.patchTranscription(mostRecentPatch)
        }
        patchQueue = []
      }, MIN_TIME_BETWEEN_PATCH)
```

Whenever a patch attempt is made, we put that patch and any subsequent patches in a queue and kick off a timer. Once the timer elapses, we take the `mostRecentPatch` (which also includes any previous patch information, LIFO here is intentional) and only send that **single** patch to be made on the backend. 

Previous patches can be disregarded since that information is included in the most recent patch. That is why the `patchQueue = []` occurs here. Once the most recent patch has occurred, we can clear the queue for newer patches and a timer will start again when a new patch enters to start the process over again.

src/store/TranscriptionsStore.js
wgranger: I'll see if it makes more sense to put this in the store model

wgranger:
Attached `patchQueue` to the store, which required creating a new function. Also realized I need to change the UI to say "SAVING..." when that is happening. Change incoming

wgranger:
@shaunanoordin Depending on what solution we find here I may remove the async test file. It's a lot of testing to cover a couple lines of code, and I'm not sure how robust the tests are.

shaunanoordin:
> Every time we save a transcription, the entire transcription is sent over to be updated: the JSON text blob, attributes, etc. Because of that, if we send two quick patches in at once, it's only imperative the most recent patch is actually applied, since changes in the first patch will be included in the second patch as well.

Ah, that explains the decision for using the LIFO pattern! OK, that makes more sense.

Question: if _only_ the _most recent Patch_ is required, do we really need an array then?

An array works for sure, but I think that might be a bit of over-design. You just want to keep track of the most recent patch. e.g.:

```
const Transcription = types.model('Transcription', {
  ...
  // DELETE: patchQueue: types.array(types.frozen()),
  nextPatch: types.frozen({}),  // single Object instead of an Array
  ...
}

  function saveMostRecentPatch() {
    const mostRecentPatch = self.nextPatch
    if (mostRecentPatch) self.patchTranscription(mostRecentPatch)
    self.nextPatch = undefined
  }

  function enqueuePatch(query) {
    if (self.nextPatch) {
      setTimeout(() => self.saveMostRecentPatch(), MIN_TIME_BETWEEN_PATCH)
    }
    self.nextPatch = query
  }
```

Either way, now that I know the intent of the code I'm happy to give this PR a 👍 Let me run one last check.

shaunanoordin:
Wait, sorry, I have to retract that 👍 and replace it with a 🤷‍♂️  because I just remembered why I said we need `self.asyncState = ASYNC_STATES.PATCHING`: the current timeout system still relies on/assumes that _all patch jobs will be complete within 3 seconds._

While the current branch (up to 826aa9c) works in normal use cases (I've tested making multiple changes like a normal human editor, and the latest change is always saved as expected) it's still possible to trigger the 409 error in a few cases:

- User's network latency is reaaally bad
- Tove database is overworked and slow to provide a response
- (what I did) did about 30+ quick changes in the editor in normal human mode (this is fine) then rapidly click on the Undo button (this isn't)

That said, the important thing here is that it works for normal human operations. If you're cool with this and are willing to accept the limitation, I'm going to approve this PR. Otherwise, we can continue looking for a more robust solution.

shaunanoordin:
Side note: I just realised that while the 409 is the root cause of the problem, there's still an additional (probably separate) fix that needs to be implemented for the `src/helpers/getError.js: getError()` function.

A 409 from Alice returns the following body...

```
{ "errors": 
  [
    { "status":"409",
      "source":null,
      "title":"Conflict",
      "detail":"Stale object error." }
  ]}
```

...and that doesn't doesn't jive with the function's expectation of `message = err.body.error`. 

In fact, I tried tracing what an error looks like when Tove returns a 409 and I see...

Screen Shot 2020-07-29 at 11 57 31

WTAF??

The error is in the form of a _readable stream_ instead of a static object. `err.json()` is a Promise, for crying out loud. How did this happen? IIRC, all other non-409 errors on Alice seem to be handled quite fine? Am I mis-remembering that?

I'm going to have to figure out what kind of errors the Alice editor can encounter to ensure the error handling system actually handles those errors - otherwise users won't be able to even see what the error messages are when they report to us.

eatyourgreens:
>The error is in the form of a readable stream instead of a static object. err.json() is a Promise, for crying out loud. How did this happen? IIRC, all other non-409 errors on Alice seem to be handled quite fine? Am I mis-remembering that?

The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) returns streams. If you're working with fetch responses directly, you need to do something like this:

```js
const response = await fetch(url)
const data = await response.body.json()
```

shaunanoordin:
Ah, that sheds some light on things - the fact that the error seems to expect some sort a Fetch pattern gives me a better idea where the error handling might have gone weird.

A typical fetch request looks like...

```
try {
  yield client.patch(`/transcriptions/${self.current.id}`, { body: query, headers: { 'If-Unmodified-Since': self.current.last_modified } }).then(response => {
    if (response.ok) {
      lastModified = getLastModified(response)
      return response
    } else {
      return Promise.reject(response)
    }
  })
  undoManager.withoutUndo(() => {
    self.error = null
    self.asyncState = ASYNC_STATES.READY
  })
} catch (err) {
  undoManager.withoutUndo(() => {
    self.error = getError(err)
    self.asyncState = ASYNC_STATES.ERROR
  })
}
```

...which is in the format...

```
// Pattern A
try {
  yield request(...)
  .then(...)
} catch (err) {
  // handle err as a static Object, not a Promise ...??? dun dun dun
}
```

With Pattern A, the error thrown by the Fetch API would be in the form a Promise (which is what I'm seeing) but the error handler expects a static Object. (i.e. the response object from the fetch action.) Basically, mixing-and-matching the try-catch pattern with the fetch action caused a hiccup.

Perhaps a better pattern is the classic Fetch/Promise one...

```
// Pattern B
yield request(...)
.then(...)
.catch(err =>
   /* handle err as a static Object, not a Promise */
)
```

Although the complication here is fitting in the `undoManager` system, which I'm only tangentially familiar with.

eatyourgreens:
That code's using a client library, which I'd expect to handle reading streams for you, returning the resolved error (or response body.)

A 409 from the server would be returned as a response body, not error, by `fetch(url)`. Again, I'd expect a client library to handle that. `fetch(url)` only errors if there's no response from the server eg. when there are network problems.

wgranger:
Yeah, I was noticing a similar thing with some of the errors. That's why there's also a [check](https://github.com/zooniverse/text-editor/blob/master/src/store/TranscriptionsStore.js#L299) for `response.ok`. If not, the response gets kicked off as an error. However, I need to double check this is working as expected in this PR.

wgranger:
Mobx state tree wants the try/catch block, or else it will throw a "can't modify state" error when trying to set state within an async call, ex: `self.asyncState`. 

I'm able to replicate the `409` issue @shaunanoordin replicated above, but only when I use the slow 3G connection. I think it's tied to the `If-Unmodified-Since` header being out of date when a PATCH is made. I believe this is what happens on a slow 3G connection:

* Remember, there are three seconds of separation between each PATCH request.
1. PATCH request is sent with the current `If-Unmodified-Since` header [set](https://github.com/zooniverse/text-editor/blob/master/src/store/TranscriptionsStore.js#L298)
2. PATCH is very slow and takes more than three seconds to return a response
3. In the meantime, a second PATCH request is fired with the _same_ `If-Unmodified-Since` headers as in step 1
4. Request one eventually resolves, but request two fails because it is using a now outdated timestamp in the `If-Unmodified-Since` header

wgranger:
#### Bug Recap
*To replicate:*
- Put connection speed on slow 3G
- Make many changes to transcription state (rearranging rows, editing lines, etc)
- Rapidly press "undo" to see some errors fail (due to `409`)

*What's happening?*
- On a 3G connection, requests are sloooooow
- Two PATCH requests occur at the same time, with the same `If-Unmodified-Since` timestamp, which TOVE will reject [here](https://github.com/zooniverse/tove/blob/974567744861da5e0e87f563784a343887f75b88/app/controllers/concerns/error_extender.rb#L58)

*How to fix?*
- Separate the time between each PATCH to more than 3 seconds (not ideal)
- Place each PATCH in a queue which will execute synchronously so each request has the newest `If-Unmodified-Since` time from the previous, completed cache (what does this implementation look like)

*Going forward*
I feel comfortable with this PR as it stands for MVP and revisiting this issue as a stretch goal. There shouldn't be any mobile users, so I'm thinking a 3G connection will be rare. Also, a subsequent successful save will clear out the `409` with the newest state.

I removed the `TranscriptionStore_Async` tests as they were brittle and hard to test with timers. I'm ok with the 0.2% drop in coverage.

PR #190, opened 2020-07-24
Fix Bug When Reordering Rows of Different Heights 

### LINKED ISSUE #189: Reordering Bug
There's a pretty bad bug in the row reorganizer that occurs when reordering two rows of different heights.

Screen Shot 2020-07-24 at 10 45 16 AM

Consider the image above. Let's say I do the following:

- Grab the second line "of this can..." and drag it over the top line "Very long text that will..."
- **Immediately** when the mouse crosses onto the top row, the two rows switch places, meaning there is a moment when the mouse is dragging "of this can..." but is hovered over "Very long text..." Illustrated below:

Screen Shot 2020-07-24 at 10 45 16 AM

- When this happens, a loop occurs. The two lines attempt to switch places constantly, sometimes over 100 times, because the different row sizes give the illusion that the mouse is constantly dragging out of one line and onto another

This isn't great UX and gives the impression that something is wrong on the site. An easy fix here is to make all lines the same height by giving a type of "very long text..." ellipsis at the end of a longer text block. However, that changes the UI a bit. 

Still investigating another way to solve this issue, but I'm not positive it can be solved simply with the drag handlers I have in place.

wgranger:
Closes #189 

Following up on a conversation in the issue above, this PR gives an ending ellipsis to long text rows. It also attached drag handlers to the transcription row text to further try to avoid the issue seen above. I believe there's an edge case where the same bug can occur when the mouse is held at the border between two rows.

Screen Shot 2020-07-24 at 11 53 52 AM

### DIFF
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -11,7 +11,7 @@ exports[`Storyshots AggregatedTranscriptions Default 1`] = `
       className="StyledBox-sc-13pk1d4-0 bdFdoP"
     >
       <div
-        className="StyledBox-sc-13pk1d4-0 dPQuSd sc-jzJRlG byHcWR"
+        className="StyledBox-sc-13pk1d4-0 dPQuSd sc-cSHVUG hDUXUW"
       >
         <div
           className="StyledBox-sc-13pk1d4-0 ikTqht"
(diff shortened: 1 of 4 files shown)


wgranger:
Screen Shot 2020-07-28 at 11 04 20 AM

@beckyrother Added the `title` tag to each row so a person can view the entire text while hovering over a line.

beckyrother:
I think this works! Will be interested to hear feedback but it does the job. Thanks!

PR #197, opened 2020-09-04
Delete Transcription Pages

### LINKED ISSUE #194: Allow Page Deletion
This was referenced in #165, but was lost in the mix. More info, including an InVision mockup, can be found on the linked comment below

https://github.com/zooniverse/text-editor/issues/165#issuecomment-653291378

wgranger:
Closes #194 

This PR allows a user to delete entire pages from a transcription. The [InVision mockup](https://projects.invisionapp.com/share/HAVICPG872T#/screens/423896487) shows that this is done with a few steps:

1. Click on the new "Delete Pages" button on the filmstrip viewer
2. Click on one of the "X" buttons that appears over a page
3. Confirm page deletion on the [modal](https://projects.invisionapp.com/share/HAVICPG872T#/screens/423896488) that appears 

When testing this, keep in mind you can undo deletion by clicking the "undo" button. I'd also like to explain the `deletePage` function (in `TranscriptionsStore`) a bit here, as there are a few different use cases that can occur through the conditional.

This PR also fixes a bug with the "Add Line" functionality. Previously, one of those added lines could appear on all slope sections for a single page. For example, if a page has a 0 degree and 180 degree section and I added a line for the 0 degree frame, that transcription would erroneously appear on both (0 and 180 degree) slope sections.

### What happens with `deletePage`?

Keep in mind, when you click the “X” to delete a page, that page and slope will also then be selected.

- First, we get the current page and slope `self.currentKey`
- remove all transcriptions from that page that match the active slope index `page.filter(line => …)`
- In Alice, a single frame `frame0` may have multiple pages since each page is defined by the frame and slope. Ex: A transcription’s `frame0` key may have transcriptions with `slope_label: 0` and `slope_label: 1`. Those would constitute two separate pages.  

#### Explaining the conditional:
- Because of this, we need to check if the current page has any more transcriptions (`page.length`) after deleting the current slope. If there are more transcriptions, we can simply reset that frame (ex: `frame0`) and reset the slope keys with `self.getSlopeKeys()`
- Else if these are the `lastTranscriptionsOnPage` (ex: `frame0` is empty and there is not a `frame0.1` or `frame0.2`) we need to remove the transcriptions but keep the `frame0` so the frontend will still be able to view that page. In the case of `frame0`, this would be the first page of the subject.
- Else we can remove this frame since the subject still has transcriptions on this page. Ex: the transcription has a `frame0` and `frame0.1`, so deleting `frame0` would be fine since we can still access the first page of the subject. To do this, we remove the frame were deleting from the `frame_order`, remove the frame from the transcription with `detach(value)` and reset the slope values with `self.getSlopeKeys()`.
- After all of the above, we set a new `self.index` and `self.slopeIndex` so we aren’t trying to access the page and slope we just deleted


mcbouslog (approved):
This LGTM, tested on http://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/TEST1/subjects/51843/edit.

In the `TranscriptionsStore` `deletePage` is handling a lot of logic. I tried to think about how some could be extracted so easier to read/understand, but I'm not sure possible, I think it's good as is.

PR #204, opened 2020-09-15
Troubleshoot Page Deletion

wgranger:
For clarification:
**frame** means the different images of a multi-image subject
**page** means the different options available via the filmstrip viewer

Some errors existed with the page deletion functionality. Including:
- When deleting the final instance of a **page** without a duplicate **frame**, the **page** should still remain active
- Copy on the Delete Page modal should be different if you're deleting the final **page** that doesn't have a **frame** duplicate
- After deleting a **page** the transcription should be saved
- Deleting the final line of a **page** should delete the entire **page** if **frame** duplicates exist
- Previously, if we deleted a **page** that didn't have any **frame** duplicates, then we deleted another **page**, that previous **page** would disappear. That should no longer happen.

I'll go ahead and stage this to [alice.preview.zooniverse.org](alice.preview.zooniverse.org) for testing

### DIFF
--- a/src/components/FilmstripViewer/FilmstripViewer.js
+++ b/src/components/FilmstripViewer/FilmstripViewer.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Box, Button, Text } from 'grommet'
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import PropTypes from 'prop-types'
 import { observer } from 'mobx-react'
 import isEqual from 'helpers/isEqual'
@@ -14,6 +14,7 @@ const RelativeBox = styled(Box)`
 
 const Uppercase = styled(Text)`
   text-transform: uppercase;
+  ${css`text-decoration: ${props => props.canDelete ? 'underline' : 'inherit'};`}
 `
 
 function usePrevious(rawValue) {
@@ -56,7 +57,7 @@ function FilmstripViewer ({
       <Box direction='row' justify='between'>
         <Text size='1em'>All Pages</Text>
         <Button
-          label={<Uppercase>Delete Pages</Uppercase>}
+          label={<Uppercase canDelete={canDelete}>Delete Pages</Uppercase>}
           onClick={() => setCanDelete(!canDelete)}
           plain
         />
--- a/src/__snapshots__/Storyshots.test.js.snap
+++ b/src/__snapshots__/Storyshots.test.js.snap
@@ -492,7 +492,7 @@ exports[`Storyshots FilmstripViewer Default 1`] = `
           type="button"
         >
           <span
-            className="StyledText-sc-1sadyjn-0 bmpSIR sc-jDwBTQ hGYXTl"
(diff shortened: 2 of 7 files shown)


snblickhan:
This all looks good to me in terms of functioning how it's meant to. 

I'd welcome some feedback here from @beckyrother in terms of how we're communicating this process to volunteers. I'd originally thought we could include some text on the 'Delete Contents' modal that says at the very bottom: "Note: you cannot delete non-duplicate images from the filmstrip viewer, only their contents." I left it out as I didn't want it to become too confusing, but now I'm wondering whether it would be helpful/informative.

Thoughts?

wgranger:
I could see the benefit of that extra explanation, but I built the dang thing and understand the complexities. It's easy to get lost with the fine distinctions between a frame and page.

beckyrother:
Agreed that some text would be helpful. Maybe: "Deleting a frame removes associated contents only; non-duplicate images cannot be removed."

Can we also just not show the X on frames that can't be removed?
I think the Delete Pages button needs an active state as well – maybe an underline when it's turned on?

wgranger:
> Agreed that some text would be helpful. Maybe: "Deleting a frame removes associated contents only; non-duplicate images cannot be removed."

I can easily add this.

> Can we also just not show the X on frames that can't be removed?

I think there will be confusion either way. A researcher will probably think "why doesn't this one page have an X at the top." I do agree it is odd to have different actions (removing an entire page vs. removing all transcriptions for that page) based on the same event (clicking the X above a page). I think I'm still leaning to having an X above all page on the filmstrip viewer. @snblickhan what do you think on this?

> I think the Delete Pages button needs an active state as well – maybe an underline when it's turned on?

Can do 👍

snblickhan:
I'm inclined to agree with you -- let's keep the X but update the pop-up text so it's clear what the action is. I'll update the About page accordingly and include an overview of this in the video tutorial as well.

wgranger:
Restaged this to alice.preview.zooniverse.org. Opening it up for a dev review now.

src/components/FilmstripViewer/FilmstripViewer.js
```diff
@@ -56,7 +57,7 @@ function FilmstripViewer ({
       <Box direction='row' justify='between'>
         <Text size='1em'>All Pages</Text>
         <Button
-          label={<Uppercase>Delete Pages</Uppercase>}
+          label={<Uppercase canDelete={canDelete}>Delete Pages</Uppercase>}
```
mcbouslog, line 60: I didn't know could pass a prop like this (`canDelete`) for styling, cool!

mcbouslog (approved):
Code looks good, fixes noted issues, 👍 . I like how the content was abstracted and use of helpers.

If I delete a page, and wait for it to complete auto-save (few seconds), then I think it takes 2 clicks of Undo to properly undo the page deletion? My guess is this is because it's undoing a save action, then the delete action? Not blocking, going to approve, but thought should be noted.

PR #205, opened 2020-10-05
ALICE About Page Edits

snblickhan:
Updating About page content for ALICE. 

- [x] Acknowledgements.js is content ONLY, needs javascript syntax/wrapping
- [x] Data Exports section is contained in DBScan.js and should be moved into its own section of About Page called DataExports.js
- [x] There's an image (screenshot of the New Task menu that shows the Transcription Task) I'd like to add in GettingStarted.js but will need assistance with syntax, etc.


srallen:
I'll have a look at this tomorrow AM.

srallen:
@snblickhan I've updated the about page to include the two new sections:

Screen Shot 2021-05-21 at 11 21 09 AM

And the new acknowledgements section in entirety:

Screen Shot 2021-05-21 at 11 21 15 AM

I have a question about an existing header for "Aggregation Settings (advanced)". On the left hand navigation, it looks like it should be styled like the other major section headers, but in the actual content it is rendered and styled like an h6:

Screen Shot 2021-05-21 at 11 21 31 AM

srallen:
Could you send me the new image to add in? I think that's all that's left.

snblickhan:
Hey! Sorry, I didn't get a notification for this, my apologies. You're right about the Aggregation style -- it should match the other major section headers.

I'll send the missing image via Slack DM!

PR #216, opened 2021-02-23
New actions

zwolf:
Run tests on creation/sync of PRs. When confirmed to work, can replace Travis.
Deploy to production on push to `production-release` tag.


.github/workflows/deploy_production.yml
```diff
@@ -0,0 +1,93 @@
...
+          creds: ${{ secrets.AZURE_STATIC_SITES }}
+
+    - name: Node.js build
+      uses: actions/setup-node@v1
+      with:
+        node-version: '14.x'
+    - run: yarn install
+    - run: yarn build
+      env:
+        REACT_APP_ENV: development
```
srallen, line 28: I think this needs to be production.
```suggestion
        REACT_APP_ENV: production
```

srallen (changes requested):
Looks good, just need the env var updated.

.github/workflows/deploy_production.yml
zwolf: I thought it had to be development so that the dev deps would be included for the build step?

srallen: This env var is just being used to set which panoptes API endpoint to use since react-scripts is controlling NODE_ENV for us and NODE_ENV is what npm will use to determine what packages to install.

zwolf: ah-hah

srallen (approved):

PR #209, opened 2021-01-21
Upgrade to react-scripts 4

shaunanoordin:
## PR Overview

This PR is a dependency update, along with a minor code change to reflect new restrictions that came from the dependency update.

Merging & deploying.

PR #238, opened 2021-09-22
Upgrade all the things

eatyourgreens:
Upgrade all packages to their latest allowed version. Upgrade `react-scripts` to 4.0.3. Remove `React` from all components (Create React App supports [the new JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#how-to-upgrade-to-the-new-jsx-transform) as of version 4.)


eatyourgreens:
This builds and runs, but the tests are completely broken. I think that it’s in the same state as #209.

Issue #233, opened 2021-08-25
Incorrect 'approved' total number on Group view page


snblickhan:
The denominator on the Group approval page always shows as being x/30 -- 30 is _not_ the total number of subjects in the group, it is the max number of subjects that can be shown on a single page. 30 here needs to show the total number of subjects in the group. 

The same error occurs in the "Download Approved Group Data" popup; the actual data export file will contain however many total subjects have been approved in a given group (even if greater than 30), but the maximum number to show up in either of these fields will be 30/30, which is misleading.

Related: if no 'Approved' subjects are present on the page being viewed, the Download button is not active. This is less urgent than the incorrect numbers showing up, but is still a bug if the button is meant to allow downloading of all approved data in a group, whether or not it's on the page being viewed. That behavior leads me to believe there was confusion in building the app around what data was being downloaded -- if it's for the whole group, it shouldn't matter, if the download was only for the subjects being viewed on a given page, it would make sense that the button is deactivated when viewing a page of unapproved subjects.

See screen recording for example: 

https://user-images.githubusercontent.com/25958358/130858807-c992f39b-f6af-46b8-a06f-cfa538f74730.mov

PR #241, opened 2021-10-28
Fix approved total number

yuenmichelle1:
Related to issue https://github.com/zooniverse/alice/issues/233
Changes made to tove (currently deployed in staging but not in prod yet), is to send over within `meta` field of the response body of `/transcriptions` call a new attribute called `approved_count` which calculates the number of approved transcriptions that come in from the api call out of all the records queries from the call. 

NOTE (to self) : Before deploying to production, must deploy tove changes to prod first. 

Screen Shot 2021-10-28 at 1 41 43 PM
Screen Shot 2021-10-28 at 1 41 51 PM

### DIFF
--- a/src/components/EditorHeader/components/Title/Title.js
+++ b/src/components/EditorHeader/components/Title/Title.js
@@ -52,7 +52,7 @@ function Title({ history, match, onEditor }) {
     header = { title: subject }
   }
   const subjectCount = header.title === group ?
-    `(${store.transcriptions.approvedCount}/${store.transcriptions.all.size} approved)` : ''
+    `(${store.transcriptions.approvedCount}/${store.transcriptions.totalCount} approved)` : ''
   return (
     <Box align='baseline'>
       <Box direction='row' wrap>
--- a/src/components/EditorHeader/components/Title/Title.spec.js
+++ b/src/components/EditorHeader/components/Title/Title.spec.js
@@ -20,7 +20,8 @@ const editContext = {
   transcriptions: {
     approvedCount: 0,
     all: { size: 0 },
-    title: 'Subject'
+    title: 'Subject',
+    totalCount: 100
   }
 }
 
@@ -71,7 +72,7 @@ describe('Component > Title', function () {
     it('should display three subheader buttons', function () {
       const approvedCount = wrapper.find(CapitalText).last().props()
       expect(wrapper.find(Button).length).toBe(3)
-      expect(approvedCount.children).toBe('(0/0 approved)')
+      expect(approvedCount.children).toBe(`(${editContext.transcriptions.approvedCount}/${editContext.transcriptions.totalCount} approved)`)
     })
 
     it('should route to the correct subheader location', function () {
--- a/src/components/Modals/DownloadDataModal/DownloadDataModalContainer.js
+++ b/src/components/Modals/DownloadDataModal/DownloadDataModalContainer.js
@@ -18,7 +18,7 @@ export default function DownloadDataModalContainer({ entireGroup }) {
       entireGroup={entireGroup}
       onClose={onClose}
       onDownload={onDownload}
-      transcriptionCount={store.transcriptions.all.size}
+      transcriptionCount={store.transcriptions.totalCount}
     />
   )
 }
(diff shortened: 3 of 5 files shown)


shaunanoordin (approved):
## PR Review

Fixes #233

This PR fixes an issue with the "Download Approved Group Data" modal showing an incorrect value for "total number of subjects in the group"

- Prior to this PR, the "DAGD" modal shows the total number of subjects PER PAGE (i.e. max 30), not the total for the whole group.
- This PR adds new variables for "approvedCount" and "totalCount", which more accurately track the numbers. I understand that these new values are coming from a recent Tove update?
- This behaviour can be seen on any "List of Subjects" page, by clicking on the "Download Approved Group Data" button.

_In the screenshot below, the "DAGD" modal shows the correct value of "63/64 SUBJECTS APPROVED", instead of "X/30 SUBJECTS APPROVED". Browsing through 3 pages of Subjects confirms that there are indeed 64 Subjects total, all but one of which are approved. [(View URL)](http://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/TEST1/subjects)_
image

### Testing

_Testing was a pain._ ☠️

Initially tested on localhost (`http://local.zooniverse.org:3000`) with macOS 11.6 + Chrome 95.
- However, all attempts to login were met with an inexplicable `invalid_grant undefined` error
- I suspect this is a combination of two things. 1. the local dev server is running on http, not httpS. (Trying to configure this for https is going to be its own thing, to be honest.)
- Chrome must have (in the past 1-2 years or so since I last accessed Alice) cranked up its security, meaning that a successful login token from `https://*.zooniverse.org` doesn't count for `http://*.zooniverse.org` . But this is just a guess.

I fell back to testing on localhost (`http://local.zooniverse.org:3000`) with macOS 11.6 + **Firefox 94,** and this actually worked!

Testing steps:
- Go to http://local.zooniverse.org:3000 and login as an admin or someone with access to projects that are being used on Alice (e.g. I'm `darkeshard`)
- From `/projects`, choose 'Anti-Slavery Testing', then 'Anti-Slavery Workflow 1', then 'Test 1' (the list should show that the Test 1 Group has 64 Subjects) [(View URL)](http://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/TEST1/subjects)
- Click on the "Download Approved Group Data" button on the top of the screen.
- The "Download Approved Group Data" modal should now pop up
  - Expected: modal should say "63/64 SUBJECTS APPROVED"
  - Actual: it does 👍

Notes:
- For comparison, on master (fa5d5134f7d5aa2a7a27333262f3b5198f4b0241) , we can see that on Test 1, the total numbers will read "29/30" on the first page, "30/30" on the second page, and "4/4" on the third page.
- The numbers are also visible next to the Group's title/name. For example:

image

### Status

Code read looks good, functionality checks out 👍 

Approved to merge, but I think we need to propagate the Tove changes to production before we deploy Alice to production too, am I correct?

yuenmichelle1:
Thanks for reviewing and thoroughness, Shaun! Much appreciated! 

> ### Testing
> _Testing was a pain._ ☠️
> 
> Initially tested on localhost (`http://local.zooniverse.org:3000`) with macOS 11.6 + Chrome 95.
> 
> * However, all attempts to login were met with an inexplicable `invalid_grant undefined` error
> * I suspect this is a combination of two things. 1. the local dev server is running on http, not httpS. (Trying to configure this for https is going to be its own thing, to be honest.)
> * Chrome must have (in the past 1-2 years or so since I last accessed Alice) cranked up its security, meaning that a successful login token from `https://*.zooniverse.org` doesn't count for `http://*.zooniverse.org` . But this is just a guess.


^ I probably should have mentioned my struggles with testing/ developing locally as well. Yes this did happen to me. 


> I fell back to testing on localhost (`http://local.zooniverse.org:3000`) with macOS 11.6 + **Firefox 94,** and this actually worked!

Yup, I had to do the same as well. Chrome had some weird issue, but I think i used Firefox and Safari to develop locally. Again, would have been helpful for me to have mentioned that to you as well. Would have saved some pain and heartache. 💔 


> ### Status
> Approved to merge, but I think we need to propagate the Tove changes to production before we deploy Alice to production too, am I correct?

Tove changes have already been deployed. So we should be good to deploy Alice to prod. 👍

PR #252, opened 2022-03-10
Include newly-required --overwrite flag in az uploads

zwolf:
MS issued a breaking change in a new version of the az client used by our CI tooling: https://docs.microsoft.com/en-us/cli/azure/release-notes-azure-cli

This PR includes the `--overwrite` flag on all `az storage blob upload-batch` commands. This change should be ported to the shared ci-cd repo and this repo migrated when this is proven to work, also.

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -40,6 +40,7 @@ jobs:
             az storage blob upload \
               --account-name zooniversestatic \
               --content-cache-control 'public, max-age=60' \
+              --overwrite
               --container-name '$web' \
               --name 'alice.zooniverse.org/commit_id.txt' \
               --file ./build/commit_id.txt
@@ -47,6 +48,7 @@ jobs:
             az storage blob upload \
               --account-name zooniversestatic \
               --content-cache-control 'public, max-age=60' \
+              --overwrite
               --container-name '$web' \
               --name 'alice.zooniverse.org/index.html' \
               --file ./build/index.html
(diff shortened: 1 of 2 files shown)


yuenmichelle1 (approved):

shaunanoordin:
Dangit! https://github.com/zooniverse/alice/actions/runs/1964794765

OK, I should try to update the [production](https://github.com/zooniverse/alice/blob/master/.github/workflows/deploy_production.yml) and [staging](https://github.com/zooniverse/alice/blob/master/.github/workflows/deploy_staging.yml) deployment scripts, I suspect these might be really old patterns.

EDIT: Wait, hang on, our last successful deploy to production was [7 days ago,](https://github.com/zooniverse/alice/actions/runs/1929057575) so the pattern can't be _that_ old and borked, even if it's still using node 14 instead of the current 16. What

shaunanoordin:
Wait, no, the checkout/login/creds pattern is solid compared to, say, [Scribes](https://github.com/zooniverse/scribes-of-the-cairo-geniza/blob/master/.github/workflows/deploy_production.yml)

...for an experiment, maybe I'll try deploying Scribes or ASM to staging.

.github/workflows/deploy_production.yml
```diff
@@ -40,20 +40,23 @@ jobs:
             az storage blob upload \
               --account-name zooniversestatic \
               --content-cache-control 'public, max-age=60' \
+              --overwrite
```
yuenmichelle1, line 43: going back to this..dont think this is the underlying issue but how important is the '\'?   @zwolf

zwolf: extremely 🤦

PR #309, opened 2022-04-03
Upgrade MobX

eatyourgreens:
Upgrade MobX to 6.7, MST to 5.1 and `mobx-react` to 7.6.

Tests are broken where they modify actions, which are immutable in MobX 6.


eatyourgreens:
This branch works when I run it locally. I can browse documents on staging, and see consensus transcriptions. I think it's only tests that are broken here, where the tests rely on implementation details of MobX that have changed in MobX 6. The editor app itself seems to work fine after the upgrade.

eatyourgreens:
Skipping all the broken tests is one way to get CI to pass, I guess.

eatyourgreens:
MobX 6.6.1 is out now, so this PR is out-of-date.

eatyourgreens:
This branch still works, but needs rebuilding with a fresh `yarn install`.

PR #313, opened 2022-04-03
Remove global variables from tests

eatyourgreens:
Test setup variables, like `wrapper`, are declared globally across all tests. This removes the global declarations and scopes test variables to the `describe` functions that use them.

I've also added a debug script so that you can run `yarn test:debug` to inspect failing tests in your favourite Node debugger.


eatyourgreens:
I _think_ Jest runs each test suite in a separate thread, which might keep those global variables separated. Don't hold me to that, though, and local function scoping seems easier to read, to me.

PR #410, opened 2023-02-03
Update UsingAlice.js

snblickhan:
Adding language to clarify that projects will not be listed in ALICE until the project data begins to retire (nb the proposed change is based on common feedback from ALICE users).


goplayoutside3 (approved):
Looks good! This is what the updated text looks like rendered on the /about page:

Screenshot 2023-02-06 at 1 39 24 PM

snblickhan:
Thanks Delilah!

PR #419, opened 2023-02-10
Redirect to /projects if already logged in

eatyourgreens:
Works around a bug where the log-in form on the home page crashes with `'id' is read-only` if you log in while already logged in.

### DIFF
--- a/src/store/AppStore.js
+++ b/src/store/AppStore.js
@@ -12,6 +12,7 @@ import { ProjectsStore } from './ProjectsStore'
 import { ModalStore } from './ModalStore'
 import { TranscriptionsStore } from './TranscriptionsStore'
 import { WorkflowsStore } from './WorkflowsStore'
+import history from '../history'
 
 const AppStore = types.model('AppStore', {
   aggregations: types.optional(AggregationsStore, () => AggregationsStore.create({})),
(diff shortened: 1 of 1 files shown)


shaunanoordin (changes requested):
## PR Review

This PR attempts to fix the issue of being unable to login on the home screen ( _on local development only,_ and only _if you're already logged in_ ) by redirecting logged-in users (when they load the app) to the Projects listing page. (i.e. what they usually see after successfully logging in.)

While this does solve the specific problem, this workaround introduces a new one: ❗ if a user attempts to directly open a URL in Alice (for example, they bookmarked [this page](https://alice.zooniverse.org/projects/4973/workflows/5339/groups/15525/subjects/13837570/edit) on their browser for future editing), they will be redirected to the Projects listing page.

This can be tested locally by - after already logging in - opening any Alice URL directly, 
e.g. https://local.zooniverse.org:3000/projects/1764/workflows/3085/groups/12345/subjects/85068/edit . (Expectation: I should be on the page editing subject 85068. Actual: I get redirected to the Projects Listing page.)

### Status

I recommend that we don't implement this particular solution.

The current login problems with Alice seems to only affect local development/testing, so we have plenty of time to work on a slightly meatier solution (e.g. home page should a "click to view Projects page!" button instead of a login form, if user's already logged in)

eatyourgreens:
Good point. We can check if you are on the home page before redirecting. That's easy. Just move the redirect down to the home page route.

PR #420, opened 2023-02-10
Build and deploy with Node 18

eatyourgreens:
- Update deprecated Node 12 actions.
- Upgrade minimum Node version.
- Run a test build as part of CI.
- Add `--openssl-legacy-provider` flag for Webpack 4 on Node 18. https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported

PR #421, opened 2023-02-10
Build with Webpack 5 on Node 18

eatyourgreens:
- upgrade Storybook to webpack 5.
- upgrade `react-scripts` to v5.
- fix tests that mock `React`.
- skip failing tests for the time being.
- includes #420, modified for Webpack 5.


eatyourgreens:
Tests are broken where they mock the store by stubbing `React.useContext`. That could be because the `React` global has been removed from React. 

It might be worth disabling the automatic runtime, just to see if that will get the tests passing. 

Longer term, Enzyme should be removed. AirBnB stopped supporting it 4 years ago, I think.

shaunanoordin (approved):
## PR Review

Supersedes #420 

This PR is a major upgrade to Alice, moving the repo to use **React Scripts**/"React Create App" v5 (which in turn uses Webpack 5 for building, among other things) _and_ Node 18 (for build/deploy pipelines and local dev).

- ⚠️ Major change: a few tests have been disabled, due to them not working with Webpack 5.
  - 👌  **I am totally OK with this.** We can fix the tests another time, focus on updating the repo would be more important.

### Testing

Tests on localhost + Chrome look good.

- Open https://local.zooniverse.org:3000/ on an _incognito_ window (otherwise, you'll run into the "already logged in" issue)
- Login as darkeshard (or other user with access to staging's Anti-Slavery Testing project)
- Go to Anti-Slavery Testing -> Anti-Slavery Workflow 1 -> Group 12345 -> (any Subject)
- Expectation: each page should render correctly with all the expected data. Final step should render the Subject Viewer.

No additional functionality tests required (e.g. Editor functionality); we're just doing a basic "does it work?" run.

### Status

LGTM 👍

PR #617, opened 2023-08-21
Revert set of commits that break transcription saving

zwolf:
Transcriptions are being saved with empty text data. This PR reverts the 18 commits that were deployed on 14 August 2023, including a major mobx version.


lcjohnso (approved):
LGTM

PR #722, opened 2023-10-31
Upgrade to Node 20

shaunanoordin:
## PR Overview

This PR updates Alice project to use Node 20.

Also see: [PFE doing the same](https://github.com/zooniverse/Panoptes-Front-End/pull/6895)


shaunanoordin:
### Testing

Tested with macOS + Chrome on localhost

- review home page & login: open `local.zooniverse.org:3000/?env=production` and log in with account `darkeshard`
  - NOTE: ?env=production is necessary, otherwise login seems to fail
- review listing pages: browse projects, then browse workflows / groups
- confirm subject viewer: from groups, select a random subject
  - confirm that subject image(s) is/are displaying, along with aggregated data (on both the subject image, and on the right-hand transcriptions panel)

Lookin' good 👍

